PR #10, opened 2026-04-03
Fix managed plugin release workflow on partner runners

lucarlig:
## Summary
- keep the full six-target release matrix for all managed plugins, including `linux-s390x` and `linux-ppc64le`
- bootstrap Python 3.12 on partner runners inside the matrixed `build-wheel` job instead of relying on `actions/setup-python`
- add regression coverage that locks `pii_filter` to the same target matrix and prevents matrix-scoped workflow logic from leaking into `preflight`

## Why
`rate-limiter-v0.0.3` failed because the new monorepo release workflow used `actions/setup-python` on `ubuntu-24.04-s390x` and `ubuntu-24.04-ppc64le`, where that action has no matching Python 3.12 manifest entry. The old successful `rate-limiter` publish workflow used system Python on those runners.

## Verification
- `python3 -m unittest tests.test_plugin_catalog.PluginCatalogTests.test_release_info_accepts_canonical_tag tests.test_plugin_catalog.PluginCatalogTests.test_release_info_gives_pii_filter_the_same_target_matrix tests.test_plugin_catalog.PluginCatalogTests.test_release_workflow_tests_artifacts_outside_source_tree`
- `python3 -m unittest tests.test_plugin_catalog tests.test_install_built_wheel`

## Review
Ran the standard detailed code review workflow in this task and addressed the actionable findings before opening this PR.

### DIFF
--- a/.github/workflows/release-rust-python-package.yaml
+++ b/.github/workflows/release-rust-python-package.yaml
@@ -91,8 +91,22 @@ jobs:
           ref: ${{ needs.resolve.outputs.checkout_ref }}
 
       - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065
+        if: ${{ matrix.runner != 'ubuntu-24.04-s390x' && matrix.runner != 'ubuntu-24.04-ppc64le' }}
         with:
           python-version: "3.12"
+
+      - name: Install system Python on partner runners
+        if: ${{ matrix.runner == 'ubuntu-24.04-s390x' || matrix.runner == 'ubuntu-24.04-ppc64le' }}
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y python3.12 python3.12-dev python3.12-venv
+          python_bin_dir="${RUNNER_TEMP}/python-bin"
+          mkdir -p "${python_bin_dir}"
+          ln -sf /usr/bin/python3.12 "${python_bin_dir}/python"
+          echo "${python_bin_dir}" >> "$GITHUB_PATH"
+          python --version
+          python -m ensurepip --upgrade
+          python -m pip install --upgrade pip
 
       - name: Verify Rust toolchain
         run: |
--- a/.github/workflows/release-rust-python-package.yaml
+++ b/.github/workflows/release-rust-python-package.yaml
@@ -132,14 +132,15 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y python3.12 python3.12-dev python3.12-venv
+          sudo apt-get clean
+          sudo rm -rf /var/lib/apt/lists/*
           python_bin_dir="${RUNNER_TEMP}/python-bin"
           mkdir -p "${python_bin_dir}"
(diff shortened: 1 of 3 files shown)


gandhipratik203 (approved):
## What's Good

- Clean root-cause fix: `actions/setup-python` has no Python 3.12 manifest for s390x/ppc64le runners, and the PR correctly bootstraps system Python on those runners instead.
- The `_release_wheel_matrix()` refactor is the right call — the old code had the full 6-target matrix only for `rate_limiter` while other plugins got 3 targets. Making all managed plugins build for all 6 targets is consistent and removes special-casing.
- Solid regression tests: the new `test_release_info_gives_pii_filter_the_same_target_matrix` locks down that `pii_filter` gets the same matrix as `rate_limiter`. The expanded workflow structural assertions in `test_release_workflow_tests_artifacts_outside_source_tree` verify the conditional logic doesn't leak into `preflight`.
- All CI checks pass, DCO signed, clean 3-commit history with logical progression.

## Findings

**F1 (low): Fragile workflow section parsing in tests**
`test_plugin_catalog.py:807-812` — The test splits the workflow YAML by literal strings like `"  preflight:\n"` and `"  build-wheel:\n"` to isolate sections. This is brittle — any indentation change, comment insertion, or job reordering will silently break the split, potentially making assertions pass vacuously on empty/wrong sections. Consider using `yaml.safe_load()` to parse the workflow and inspect job-level keys structurally, or at minimum assert the split produced non-empty sections.

**F2 (low): No `apt-get clean` after install**
`.github/workflows/release-rust-python-package.yaml:131-137` — The partner-runner bootstrap installs packages via `apt-get` but doesn't clean up the apt cache. On ephemeral GitHub runners this is cosmetic, but adding `sudo apt-get clean && sudo rm -rf /var/lib/apt/lists/*` is standard CI hygiene and can shave a few seconds off artifact caching if the runner image is cached.

**F3 (info): `ensurepip` then `pip install --upgrade pip` is redundant**
Line 138-139 — `ensurepip --upgrade` already installs the latest bundled pip. The subsequent `pip install --upgrade pip` fetches the latest from PyPI. This is fine if you explicitly want the newest pip, but if not, one of the two lines can be dropped.

**F4 (info): Symlink approach assumes `/usr/bin/python3.12` path**
Line 134 — `ln -sf /usr/bin/python3.12` assumes the Ubuntu package installs the binary at that exact path. This is correct for `ubuntu-24.04` today, but if the runner image changes, it could break. A more robust alternative: `ln -sf "$(which python3.12)" "${python_bin_dir}/python"`. Minor since these are GitHub-managed runner images with predictable layouts.

## Verdict

Clean, well-motivated fix with good test coverage. The 6-target matrix unification and conditional Python bootstrap are correct. All findings are low/informational — nothing blocks merge. Ship it.

lucarlig:
Addressed in `bebade6`.

- Made the workflow test section checks less brittle.
- Cleaned apt metadata, switched the symlink to `$(which python3.12)`, and dropped the redundant pip upgrade.

Re-ran the focused workflow tests, `plugin_catalog.py validate .`, and the full Python unittest suite.

PR #13, opened 2026-04-07
Add explicit plugin kind metadata

### LINKED ISSUE #12: Plugin classname which implements available hooks is not detailed in plugin-manifest or project files.
While implementing feature [Plugin bundling, catalog, installation and versioning](https://github.com/contextforge-org/contextforge-plugins-framework/issues/9) for the contextforge-plugins-framework, I noticed that the plugin classname which implements the available hooks is not available in any of the pyproject.toml, Cargo.toml, or plugin-manifest.yaml files.

Would it be possible to provide the plugin's class that provides the hooks in the plugin-manifest.yaml file?

e.g. for the `cpex-pii-filter` package, the updated plugin-manifest.yaml could contain a `kind` name value pair:

```yaml
kind: cpex_pii_filter.pii_filter.PIIFilterPlugin
   .
   .
   .
```

lucarlig:
## Summary
- require each plugin manifest to declare an explicit `kind`
- publish the same plugin class path through `[project.entry-points."cpex.plugins"]`
- expose and validate `kind` in `tools/plugin_catalog.py`, with tests and docs updated accordingly

## Why
Issue #12 identified that PyPI package installation is deterministic, but plugin class discovery was still implicit. This change makes the plugin class path an explicit repository contract while also exposing a standard Python entry point for runtime discovery.

Closes #12

## Validation
- `python3 -m unittest tests.test_install_built_wheel`

### DIFF
--- a/plugins/rust/python-package/pii_filter/cpex_pii_filter/plugin-manifest.yaml
+++ b/plugins/rust/python-package/pii_filter/cpex_pii_filter/plugin-manifest.yaml
@@ -1,7 +1,7 @@
 description: "Rust-backed PII detection and masking for prompt arguments, tool inputs, and tool outputs"
 author: "ContextForge Contributors"
 version: "0.2.0"
-kind: "cpex_pii_filter.pii_filter:PIIFilterPlugin"
+kind: "cpex_pii_filter.pii_filter.PIIFilterPlugin"
 available_hooks:
   - "prompt_pre_fetch"
   - "prompt_post_fetch"
--- a/plugins/rust/python-package/pii_filter/cpex_pii_filter/plugin-manifest.yaml
+++ b/plugins/rust/python-package/pii_filter/cpex_pii_filter/plugin-manifest.yaml
@@ -1,6 +1,7 @@
 description: "Rust-backed PII detection and masking for prompt arguments, tool inputs, and tool outputs"
 author: "ContextForge Contributors"
 version: "0.2.0"
+kind: "cpex_pii_filter.pii_filter:PIIFilterPlugin"
 available_hooks:
   - "prompt_pre_fetch"
   - "prompt_post_fetch"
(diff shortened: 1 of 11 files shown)


lucarlig:
@tedhabeck please check this would close the issue correctly

plugins/rust/python-package/pii_filter/cpex_pii_filter/plugin-manifest.yaml
```diff
@@ -1,6 +1,7 @@
 description: "Rust-backed PII detection and masking for prompt arguments, tool inputs, and tool outputs"
 author: "ContextForge Contributors"
 version: "0.2.0"
+kind: "cpex_pii_filter.pii_filter:PIIFilterPlugin"
```
tedhabeck: Minor issue, but I'm curious about why a semicolon, rather than a dot (period)?  The main branch of the CPEX framework plugin loader expects a period.  See:

- https://github.com/contextforge-org/contextforge-plugins-framework/blob/main/cpex/framework/loader/plugin.py#L70. 
- https://github.com/contextforge-org/contextforge-plugins-framework/blob/main/cpex/framework/utils.py#L186

It make sense for the pyproject.toml as it's consistent with the entry-points data model:
- https://packaging.python.org/en/latest/specifications/entry-points/#entry-points

I can work with it either way, it's a minor update.

plugins/rust/python-package/rate_limiter/cpex_rate_limiter/plugin-manifest.yaml
```diff
@@ -1,6 +1,7 @@
 description: "Rate limiting by user/tenant/tool — memory (single-process) or Redis (shared across instances)"
 author: "ContextForge Contributors"
 version: "0.0.3"
+kind: "cpex_rate_limiter.rate_limiter:RateLimiterPlugin"
```
tedhabeck: ibid

tedhabeck (approved):
LGTM 👍

plugins/rust/python-package/pii_filter/cpex_pii_filter/plugin-manifest.yaml
lucarlig: either is aligned with `pyproject.toml` or with CPEX framework. i switched to `.`

PR #25, opened 2026-04-10
fix: add missing Rust CI parity checks

lucarlig:
## Summary
- add Rust plugin CI parity jobs for workspace cargo-deny, Rust coverage, and Rust documentation
- replace `cargo audit` with `cargo deny check --config deny.toml`
- consolidate plugin `deny.toml` files into one root policy and keep known `unic-*` unmaintained advisories as explicit scoped ignores
- trigger/select all plugins for root dependency-policy inputs such as `deny.toml` and `Cargo.lock`
- run coverage/docs against selected Cargo packages for single- and multi-plugin changes
- enforce a 50.00% per-plugin line coverage floor with catalog-aware Cobertura validation
- validate workflow dispatch as an all-plugin run and validate workflow outputs before downstream shell use
- update `SECURITY.md` and generated PII filter stubs to match the centralized CI policy

## Validation
- `python3 -m unittest tests/test_plugin_catalog.py`
- `cargo deny check --config deny.toml`
- `cargo check --workspace`
- `actionlint .github/workflows/ci-rust-python-package.yaml`
- `cargo llvm-cov -p encoded_exfil_detection -p pii_filter -p rate_limiter -p retry_with_backoff -p secrets_detection -p url_reputation --cobertura --output-path coverage/cobertura.xml`
- `python3 tools/plugin_catalog.py coverage-check . coverage/cobertura.xml 50.00 '["encoded_exfil_detection","pii_filter","rate_limiter","retry_with_backoff","secrets_detection","url_reputation"]'`

## Review Notes
- Latest detailed review pass had no actionable correctness, testing, docs/API, performance, or architecture findings.
- A final security finding about PR-controlled JSON flowing into shell interpolation was fixed by static output validation plus env-based JSON parsing in downstream jobs.
- Remaining cargo-deny ignores are scoped to the current transitive `pyo3-stub-gen` / `rustpython-parser` `unic-*` advisories, so new unmaintained advisories still fail CI.


gandhipratik203:
## What’s Good

- Nice PR overall. Pulling the Rust plugin CI closer to real parity by adding dedicated security, coverage, and documentation jobs makes the repo-level contract much clearer than relying only on per-plugin `make ci`.
- The coverage work feels thoughtfully done. Passing explicit `cargo_packages` through the workflow and tightening `tools/plugin_catalog.py` so `coverage-check` validates the expected plugin set makes the 50% per-plugin floor feel intentional rather than best-effort.
- Consolidating the plugin-local `deny.toml` files into a single workspace policy is the right maintenance direction, and I like that the PR also updates the surrounding tests and `SECURITY.md` so the repo’s behavior and docs stay aligned.
- The smaller cleanup pieces are good too: stabilizing the generated `pii_filter` stubs should reduce CI churn, and adding `bench-no-run` to `retry_with_backoff` brings that plugin closer to the same CI shape as the others.

## Findings

### [Medium] Workspace `deny.toml` changes still bypass the new security-policy path

This PR makes the workspace `deny.toml` the source of truth, but `.github/workflows/ci-rust-python-package.yaml` still does not include root `deny.toml` in the `push` / `pull_request` path filters, and `tools/plugin_catalog.py` still does not treat `deny.toml` as a shared-path trigger. As a result, a follow-up PR that only adjusts the workspace cargo-deny policy will either skip this workflow entirely or select no plugins and skip `security-policy`, so the policy-only change is not actually being validated by the new `cargo deny` job.

gandhipratik203 (approved):
Looks good to me. Thanks for tightening the root deny.toml trigger path and adding the regression coverage around the shared-path selection and workflow contract.

PR #74, opened 2026-05-01
fix(rate-limiter): production hardening — TLS support for managed Redis

gandhipratik203:
## Summary

Enables TLS for the rate limiter's Redis backend so deployments against managed Redis services with in-transit encryption (AWS ElastiCache, Redis Cloud, etc.) work end-to-end. Bumps `cpex_rate_limiter` 0.0.4 → 0.0.6.

---

## Gaps closed

**Gap A (HIGH)** — The Redis backend cannot connect to managed Redis services that require TLS. Operators using AWS ElastiCache with in-transit encryption (or any other `rediss://` URL) can configure the binding successfully — the binding API returns HTTP 200 — but the plugin's backend init raises `InvalidClientConfig: can't connect with TLS, the feature is not enabled`. The plugin manager logs `Failed to load plugin RateLimiterPlugin` and silently skips it under the default `fail_on_plugin_error: false` setting, leaving rate limits unenforced while the operator believes them to be in place. Fixed by compiling the redis crate with rustls-based TLS features.

**Gap B (MEDIUM)** — The redis crate version previously pinned (0.27) carried `rustls-pemfile` 2.2.0 transitively, which `cargo-deny` flagged as unmaintained (RUSTSEC-2025-0134) and blocked CI's `security-policy` stage. Fixed by bumping the redis crate to 0.32, where the rustls-native-certs upgrade dropped the `rustls-pemfile` dep entirely.

**Gap C (MEDIUM)** — rustls 0.23 dropped its implicit default crypto provider; the redis crate's `tls-rustls` feature does not pick one. Without an explicit install, the first TLS handshake panicked with `Call CryptoProvider::install_default() before this point...`. The Python shim's broad except caught the panic and `fail_mode=open` allowed the request through, so deployments would still appear to load the plugin while quietly enforcing nothing — same observable end state as Gap A, different log message. Fixed by installing the `ring` crypto provider once at plugin init via a `OnceLock`-guarded call in `RateLimiterPluginCore::new()`.

---

## Test results

<details>

```text
cargo test --lib                      57 passed
uv run pytest tests/rate_limiter/     107 passed
```

</details>

<details>
<summary>New tests added in this PR</summary>

**`TestRedisTlsSupport`** — regression coverage for managed-Redis TLS support

| Test | Target |
|---|---|
| `test_rediss_url_does_not_fail_with_tls_not_enabled` | Constructs the plugin with a `rediss://` URL; asserts no `InvalidClientConfig` is raised at backend init. Pins Gap A. |
| `test_rediss_url_lazy_handshake_reaches_connectivity_layer` | Drives a `tool_pre_invoke` against a closed `rediss://` port so the lazy TLS handshake actually engages; asserts the captured warning does not contain the TLS-feature-not-enabled signature (`feature is not enabled` / `InvalidClientConfig`). Pins Gap A and Gap C jointly. |

</details>

---

## Migration notes

- **No new required config.** Existing deployments using `redis://` URLs continue to work unchanged. `rediss://` URLs now work where they previously failed.
- **CA bundle expectation.** The plugin reads CA roots from the host OS trust store at runtime via `rustls-native-certs`. Container images that omit `ca-certificates` (some distroless / scratch / unconfigured Alpine variants) would need to install the bundle before TLS endpoints can be verified. UBI Minimal — what mcp-context-forge ships on — already includes the CA bundle by default; AWS ElastiCache uses public-CA chains that are present there.

---

## Follow-ups

- Tighten the TLS lazy-handshake test's positive assertion (currently scoped to the TLS-feature-not-enabled negative signature only). Would require the Rust core's `log_exception()` to surface the underlying `RedisError` text instead of the generic "error; allowing request" message — the error details are dropped at the Python boundary today.
- Real-TLS-Redis fixture for end-to-end handshake verification (self-signed cert + CA-pinning fixture). Heavier than the connectivity-shape variant; deferred until needed.
- Companion main-repo PR to bump the `cpex-rate-limiter` pin to 0.0.5 once published to PyPI.
- Wipe-on-disable feature (counter cleanup when an operator transitions the plugin to `mode=disabled`) and the round-1/round-2 review feedback addressing it have been moved to a separate branch (`feat/rate-limiter-wipe-on-disable-only`); a follow-up PR will be linked here once opened.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -63,7 +63,6 @@ allow = [
     "CC0-1.0",
     "CDDL-1.0",
     "CDDL-1.1",
-    "CDLA-Permissive-2.0",
     "ClArtistic",
     "CPL-1.0",
     "curl",
--- a/deny.toml
+++ b/deny.toml
@@ -63,6 +63,7 @@ allow = [
     "CC0-1.0",
     "CDDL-1.0",
     "CDDL-1.1",
+    "CDLA-Permissive-2.0",
     "ClArtistic",
     "CPL-1.0",
     "curl",
--- a/plugins/tests/rate_limiter/test_redis_integration.py
+++ b/plugins/tests/rate_limiter/test_redis_integration.py
@@ -1578,3 +1578,42 @@ class TestConfigHardening:
             "engine must warn on unknown config keys so misspellings are visible — "
             f"captured records: {[(r.levelname, r.getMessage()) for r in caplog.records]}"
         )
+
+
+class TestRedisTlsSupport:
+    """TLS / rediss:// scheme support.
+
+    Regression coverage for wo-tracker #68217: managed Redis services
+    (AWS ElastiCache with in-transit encryption, Redis Cloud, etc.) require
+    a `rediss://` URL. The Rust engine must be built with the redis crate's
+    TLS feature so that constructing a client with a `rediss://` URL parses
+    successfully — without it the engine raises `InvalidClientConfig: can't
+    connect with TLS, the feature is not enabled` at plugin init and the
+    plugin is silently skipped.
+    """
+
+    @pytest.mark.asyncio
+    async def test_rediss_url_does_not_fail_with_tls_not_enabled(self):
+        """Constructing the plugin with a rediss:// URL must not raise InvalidClientConfig.
+
+        Construction is independent of connectivity: the redis crate parses
+        the URL and creates a lazy client. We point at a port nothing is
+        listening on so the test never opens a real TLS handshake.
+        """
+        # rediss:// = TLS scheme. Port 1 has no listener; we never attempt
+        # a real connection here — the regression is at URL/feature parsing.
+        plugin = RateLimiterPlugin(
+            PluginConfig(
+                name="RateLimiter",
+                kind="cpex_rate_limiter.rate_limiter.RateLimiterPlugin",
+                hooks=["tool_pre_invoke"],
+                priority=100,
+                config={
+                    "by_user": "3/s",
+                    "backend": "redis",
+                    "redis_url": "rediss://127.0.0.1:1/15",
+                    "algorithm": "fixed_window",
+                },
+            )
+        )
+        assert plugin is not None
--- a/plugins/tests/rate_limiter/test_redis_integration.py
+++ b/plugins/tests/rate_limiter/test_redis_integration.py
@@ -1684,3 +1684,319 @@ class TestRedisTlsSupport:
         )
 
 
+# =============================================================================
+# Real TLS Redis handshake test
+# =============================================================================
+# The TestRedisTlsSupport class above only verifies that the redis crate was
+# compiled with TLS support and that the wo-tracker #68217 InvalidClientConfig
+# signature does not appear — neither test drives a real TLS handshake.
+# A handshake regression (missing crypto provider, bad cert chain, broken
+# rustls-native-certs lookup) could pass those tests while failing in
+# production.
+#
+# This fixture spins up a Redis container listening on TLS only, behind a
+# self-signed CA generated fresh in /tmp.  SSL_CERT_FILE is exported so both
+# rustls-native-certs (the Rust core's TLS path) and Python's ssl module
+# (test-side helpers) trust the test CA without touching the OS trust store.
+# The container is module-scoped and torn down on teardown.
+#
+# The whole stack skips cleanly when openssl or docker are unavailable, so
+# laptops without docker stay green while Linux CI runners exercise the full
+# end-to-end path.
+# =============================================================================
+
+
+_TEST_TLS_REDIS_CONTAINER_PORT = 16390
+_TEST_TLS_REDIS_CONTAINER_NAME = "pytest-rl-redis-tls-integ"
+
+
+def _openssl_available() -> bool:
+    """Return True if the openssl CLI is on PATH."""
+    try:
+        result = subprocess.run(
+            ["openssl", "version"],
+            check=False,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+            timeout=5,
+        )
+        return result.returncode == 0
+    except (FileNotFoundError, subprocess.TimeoutExpired):
+        return False
+
+
+def _docker_available() -> bool:
+    """Return True if the docker CLI is on PATH and the daemon is reachable."""
+    try:
+        result = subprocess.run(
+            ["docker", "version", "--format", "{{.Server.Version}}"],
+            check=False,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+            timeout=5,
+        )
+        return result.returncode == 0
+    except (FileNotFoundError, subprocess.TimeoutExpired):
+        return False
+
+
+def _generate_self_signed_redis_cert(cert_dir: str) -> tuple[str, str, str]:
+    """Generate a self-signed CA and a Redis server cert in cert_dir.
+
+    Mirrors mcp-context-forge/tls-test/gen-certs.sh but with a 1-day TTL
+    (long enough to never expire mid-test, short enough to make leakage
+    obvious) and 2048-bit RSA (faster than 4096; ephemeral test material
+    only).  Returns (ca.crt, redis.crt, redis.key) as absolute paths.
+    """
+    ca_key = os.path.join(cert_dir, "ca.key")
+    ca_crt = os.path.join(cert_dir, "ca.crt")
+    redis_key = os.path.join(cert_dir, "redis.key")
+    redis_crt = os.path.join(cert_dir, "redis.crt")
+    redis_csr = os.path.join(cert_dir, "redis.csr")
+    redis_ext = os.path.join(cert_dir, "redis.ext")
+
+    def _ssl(*args: str) -> None:
+        subprocess.run(
+            ["openssl", *args],
+            check=True,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.PIPE,
+        )
+
+    # Self-signed CA.  -addext is required for OpenSSL 3.x clients,
+    # which refuse to verify chains where the CA cert lacks an
+    # explicit keyUsage extension ("CA cert does not include key
+    # usage extension").  Older clients ignore the extra extensions.
+    _ssl("genrsa", "-out", ca_key, "2048")
+    _ssl(
+        "req", "-x509", "-new", "-nodes",
+        "-key", ca_key, "-sha256", "-days", "1",
+        "-subj", "/CN=cpex-pytest-tls-ca",
+        "-addext", "basicConstraints=critical,CA:true",
+        "-addext", "keyUsage=critical,keyCertSign,cRLSign",
+        "-out", ca_crt,
+    )
+
+    # Redis server cert signed by the CA.  SAN covers localhost +
+    # 127.0.0.1 so both the in-test PING and the plugin's connection
+    # to the host-mapped port verify cleanly.
+    _ssl("genrsa", "-out", redis_key, "2048")
+    _ssl("req", "-new", "-key", redis_key, "-subj", "/CN=localhost", "-out", redis_csr)
+    with open(redis_ext, "w") as f:
+        f.write("subjectAltName = DNS:localhost,IP:127.0.0.1\nextendedKeyUsage = serverAuth\n")
+    _ssl(
+        "x509", "-req", "-in", redis_csr,
+        "-CA", ca_crt, "-CAkey", ca_key, "-CAcreateserial",
+        "-out", redis_crt, "-days", "1", "-sha256",
+        "-extfile", redis_ext,
+    )
+
+    # Permissions: world-readable cert/key files so the redis user
+    # inside the container (typically UID 999) can read them through
+    # the bind mount; keys are still locked down to the file owner
+    # for write.
+    for p in (ca_crt, redis_crt, redis_key):
+        os.chmod(p, 0o644)
+
+    return ca_crt, redis_crt, redis_key
+
+
+def _tls_redis_pings(host: str, port: int, ca_crt: str, timeout: float = 1.0) -> bool:
+    """Return True if a TLS PING to host:port verifies against ca_crt and returns PONG."""
+    try:
+        # Third-Party
+        import redis as _redis_sync  # noqa: PLC0415
+    except Exception:
+        return False
+    try:
+        client = _redis_sync.Redis(
+            host=host,
+            port=port,
+            ssl=True,
+            ssl_ca_certs=ca_crt,
+            socket_connect_timeout=timeout,
+            socket_timeout=timeout,
+        )
+        try:
+            return bool(client.ping())
+        finally:
+            try:
+                client.close()
+            except Exception:
+                pass
+    except Exception:
+        return False
+
+
+@pytest.fixture(scope="module")
+def redis_tls_url_for_integration():
+    """Yield a rediss:// URL backed by a real TLS-enabled Redis container.
+
+    Closes the gap left by TestRedisTlsSupport (which is construction-only).
+    Generates a fresh self-signed CA + Redis cert in /tmp, starts a
+    redis:7 container with --tls-port 6390 (host port 16390), and exports
+    SSL_CERT_FILE so both rustls-native-certs (Rust core) and Python's
+    ssl module (test-side helpers) pick up the test CA.
+
+    Skips cleanly if openssl or docker is unavailable.  An external
+    operator can short-circuit the cert/container setup by exporting
+    RATE_LIMITER_TLS_REDIS_URL=rediss://<host>:<port>/15 with a CA the
+    current process already trusts.
+    """
+    override = os.environ.get("RATE_LIMITER_TLS_REDIS_URL")
+    if override:
+        # Caller-managed endpoint and trust store; we leave SSL_CERT_FILE alone.
+        yield override
+        return
+
+    try:
+        # Third-Party
+        import redis.asyncio  # noqa: F401, PLC0415
+    except Exception:
+        pytest.skip("redis.asyncio package not installed")
+
+    if not _openssl_available():
+        pytest.skip("openssl CLI not available; skipping real TLS handshake test")
+    if not _docker_available():
+        pytest.skip("docker not available; skipping real TLS handshake test")
+
+    # The cert dir has two constraints: pytest's tmp_path_factory parents
+    # are mode 0700 (the redis container's UID 999 cannot traverse them),
+    # and /tmp is not always shared into the container runtime VM
+    # (Docker Desktop / colima on macOS share $HOME but not /tmp). A
+    # subdir under $HOME satisfies both: traversable on Linux CI and
+    # bind-mountable on dev laptops.
+    # Standard
+    import shutil  # noqa: PLC0415
+    import tempfile  # noqa: PLC0415
+
+    cache_root = os.path.join(os.path.expanduser("~"), ".cache", "cpex-pytest-tls")
+    os.makedirs(cache_root, exist_ok=True)
+    cert_dir = tempfile.mkdtemp(prefix="cpex-tls-redis-", dir=cache_root)
+    os.chmod(cert_dir, 0o755)
+
+    try:
+        ca_crt, _redis_crt, _redis_key = _generate_self_signed_redis_cert(cert_dir)
+    except subprocess.CalledProcessError as exc:
+        shutil.rmtree(cert_dir, ignore_errors=True)
+        pytest.skip(f"openssl failed to generate test certs: {exc.stderr!r}")
+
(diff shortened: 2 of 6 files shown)


plugins/rust/python-package/rate_limiter/cpex_rate_limiter/rate_limiter.py
```diff
@@ -74,16 +74,123 @@ async def initialize(self) -> None:
...
+        "disabled" via the admin mode API, every rate-limit counter for this
+        plugin's configured key prefix is deleted from Redis before shutdown
+        completes. Re-enabling then starts every user with a fresh window.
+        See README "Disabling resets counters" for the contract details and
+        the binding-API limitation.
         """
         try:
-            self._core.shutdown()
+            if await self._mode_in_redis_says_disabled():
+                await self._wipe_my_counters()
```
lucarlig: [P2] Every instance that sees mode `disabled` runs the full namespace scan and delete. In a fleet, disable becomes roughly N workers times M keys, with duplicate deletes. Use a Redis single-flight guard such as `SET ... NX EX` so only one worker performs the wipe.

plugins/rust/python-package/rate_limiter/cpex_rate_limiter/rate_limiter.py
```diff
@@ -74,16 +74,123 @@ async def initialize(self) -> None:
...
+        try:
+            batch: list = []
+            deleted = 0
+            async for key in client.scan_iter(match=pattern, count=500):
+                batch.append(key)
+                if len(batch) >= 500:
+                    deleted += await client.delete(*batch)
+                    batch.clear()
+            if batch:
+                deleted += await client.delete(*batch)
```
lucarlig: [P2] `SCAN` avoids blocking key enumeration, but `DEL` still frees values on Redis' main thread. For large counter sets, especially sorted-set sliding-window counters, 500-key batches can spike shared Redis latency. Prefer `UNLINK` with a `DEL` fallback.

plugins/rust/python-package/rate_limiter/cpex_rate_limiter/rate_limiter.py
```diff
@@ -74,16 +74,123 @@ async def initialize(self) -> None:
...
+        plugin instance is garbage-collected.
+
+        Wipe behaviour: if the operator has just set this plugin's mode to
+        "disabled" via the admin mode API, every rate-limit counter for this
+        plugin's configured key prefix is deleted from Redis before shutdown
+        completes. Re-enabling then starts every user with a fresh window.
+        See README "Disabling resets counters" for the contract details and
+        the binding-API limitation.
         """
         try:
```
lucarlig, line 83: [P2] `shutdown()` infers disable by reading `plugin:{name}:mode` from Redis, then deletes backend keys from the Python shim. This couples plugin lifecycle, gateway admin storage, pub/sub ordering, and Redis key schema. Safer shape: framework passes explicit shutdown reason/target mode, and the Rust backend owns deletion because it owns key construction.

lucarlig (commented):
Detailed review pass found no Critical issues. I left five P2 findings covering Redis wipe safety/performance, lifecycle coupling, and test coverage gaps.

plugins/tests/rate_limiter/test_redis_integration.py
```diff
@@ -1578,3 +1578,221 @@ async def test_unknown_config_key_emits_warning(self, redis_url_for_integration,
...
+                priority=100,
+                config={
+                    "by_user": "3/s",
+                    "backend": "redis",
+                    "redis_url": "rediss://127.0.0.1:1/15",
+                    "algorithm": "fixed_window",
+                },
+            )
+        )
+        assert plugin is not None
```
lucarlig, line 1619: [P2] The TLS regression test only constructs a plugin with `rediss://127.0.0.1:1/15`; it does not drive `tool_pre_invoke` or a real TLS Redis operation. That can miss lazy connection, connector, roots, or runtime feature failures. Add a TLS Redis fixture, or at least assert the lazy path fails as connectivity rather than TLS feature/config.

plugins/tests/rate_limiter/test_redis_integration.py
```diff
@@ -1578,3 +1578,221 @@ async def test_unknown_config_key_emits_warning(self, redis_url_for_integration,
...
+
+async def _seed_rate_limit_counters(redis_url: str, count: int = 3) -> None:
+    """Drop ``count`` rl:* keys into Redis to simulate accumulated counter state."""
+    # Third-Party
+    import redis.asyncio as aioredis  # noqa: PLC0415
+
+    client = aioredis.from_url(redis_url, decode_responses=True)
+    try:
+        for i in range(count):
+            await client.set(f"rl:test:user:user-{i}:60", "5", ex=60)
```
lucarlig: [P2] Tests seed only hard-coded `rl:*` keys and never configure `redis_key_prefix`. A regression that ignores custom prefixes, wipes only defaults, or deletes adjacent namespaces could pass. Add a case with matching and non-matching custom-prefix keys and assert only the configured prefix is deleted.

msureshkumar88:
## Code Review — PR #74: production hardening (TLS + wipe-on-disable)

### Overview

Closes two real gaps in the Redis path:
- **Gap A (HIGH)**: `rediss://` URLs blow up at plugin init because the redis crate was compiled without TLS features. Fixed by enabling `tls-rustls*` + `tokio-rustls-comp`.
- **Gap B (MEDIUM)**: Counter state survives a disable→re-enable cycle, causing unexpected immediate throttling. Fixed by reading `plugin:<name>:mode` from Redis at shutdown and wiping `<prefix>:*` when the value is `"disabled"`.

Architecture is sound and the wipe contract (conditional, fail-safe, idempotent, non-blocking on core shutdown) is correctly implemented.

---

### Blocking Issues

#### 1. `security-policy` CI failure — unmaintained `rustls-pemfile` (RUSTSEC-2025-0134)

`cargo-deny` rejects the build because `tls-rustls` pulls in `rustls-native-certs` → `rustls-pemfile`, which was archived in August 2025 with no safe upgrade path.

The `tls-rustls` feature is redundant here. The PR's stated goal ("managed services with public-CA certificates work without further configuration") is fully satisfied by `tls-rustls-webpki-roots` alone — it bundles Mozilla CA roots and doesn't touch `rustls-native-certs`.

**Fix:** drop `tls-rustls` from `Cargo.toml`:

```toml
# before
redis = { version = "0.27", features = ["aio", "tokio-comp", "tls-rustls", "tls-rustls-webpki-roots", "tokio-rustls-comp"] }

# after — removes rustls-native-certs → rustls-pemfile from the dep tree
redis = { version = "0.27", features = ["aio", "tokio-comp", "tls-rustls-webpki-roots", "tokio-rustls-comp"] }
```

#### 2. `mutation-testing (rate_limiter)` CI failure — stale diff baseline

`cargo-mutants` errors with:

```
diff has:   "        Python::attach( |py| -> PyResult<()> {"
source has: "        // Ensure the embedded interpreter is initialized for this test process."
The diff might be out of date with this source tree.
```

The action computes the diff between `BASE_SHA` and `HEAD_SHA`. If the branch was rebased after the run started, or if `BASE_SHA` points to a commit whose tree no longer matches the working tree, `cargo-mutants` rejects the diff. A fresh push after the `rustls-pemfile` fix above should regenerate a consistent baseline. If it persists, verify `BASE_SHA` in the CI workflow resolves to the correct merge-base.

(`mutation-testing (secrets_detection)` failure appears unrelated to this PR — no secrets_detection files changed.)

---

### Non-Blocking Observations

**Python connection overhead on shutdown**

`_mode_in_redis_says_disabled()` opens a brand-new `redis.asyncio` client, issues one `GET`, then closes it. The Rust core holds an already-open connection, but it's not reachable from Python at shutdown time. This is fine in practice (shutdown is rare), but worth a brief inline comment explaining why a new client is opened instead of reusing the Rust one.

**Type annotation**

```python
batch: list = []
```

Should be `list[str | bytes]` (or whatever `aioredis.scan_iter` yields) to keep the type checker happy and signal intent.

**Deferred import**

```python
try:
    import redis.asyncio as aioredis  # noqa: PLC0415
except Exception:
    return False
```

`redis` is now a declared dep in `pyproject.toml`, so the `ImportError` guard is dead code. A top-level import is cleaner. The `PLC0415` suppression hint can also be dropped.

**Batch size magic number**

The 500-key batch in `_wipe_my_counters()` is reasonable but uncommented. Worth a one-liner: `# 500 keys per DEL keeps round-trip count low without risking a single oversized command`.

---

### Test Coverage

Good. The six `TestWipeOnDisable` cases cover the full conditional matrix cleanly, and the concurrent-shutdown test is exactly the right thing to have. One gap worth considering in a follow-up: a test that asserts `SCAN` is used instead of `KEYS` (mock the redis client and verify `scan_iter` was called, not `keys`). Not a blocker for this PR since the implementation visibly uses `scan_iter`, but it would prevent regression if someone refactors the wipe path.

---

### Summary

**Two blockers before merge:**
1. Remove `tls-rustls` feature to eliminate the `rustls-pemfile` advisory and unblock `security-policy`.
2. Confirm mutation-testing passes after a fresh push (likely resolves itself once (1) is fixed and a new run is triggered with a fresh diff baseline).

Code quality is good. The wipe logic is correct, the fail-safe properties are well-reasoned, and the README/migration notes are clear. Approve pending the two CI fixes.

plugins/rust/python-package/rate_limiter/cpex_rate_limiter/rate_limiter.py
gandhipratik203: Thanks — addressed in 2b453b6. Added `_acquire_wipe_lock()` / `_release_wipe_lock()` using `SET <key> 1 NX EX 30`. Lock key is `<prefix>-wipe-lock` (dash separator) so the wipe's `<prefix>:*` SCAN doesn't match it. Released eagerly after the wipe so the next disable cycle is responsive; the TTL is the safety net if the lock holder dies mid-wipe.

Test: `test_wipe_skipped_when_another_worker_holds_lock` pre-acquires the lock externally, runs `shutdown()` with `mode=disabled`, and asserts the counters survive — proving the wipe path was skipped.

plugins/rust/python-package/rate_limiter/cpex_rate_limiter/rate_limiter.py
gandhipratik203: Thanks — addressed in 2b453b6. Switched the per-batch drop to `UNLINK` with a `ResponseError → DEL` fallback so pre-Redis-4.0 deployments still work. Existing `TestWipeOnDisable` cases pass unchanged (the swap is functionally transparent on modern Redis).

plugins/rust/python-package/rate_limiter/cpex_rate_limiter/rate_limiter.py
gandhipratik203: Agreed this is the right long-term shape — captured as a follow-up in the PR description (Follow-ups list, fourth bullet). Out of scope for this sev-1 fix because it requires a framework-API addition in mcp-context-forge plus moving wipe into the Rust core; deserves a focused review.

plugins/tests/rate_limiter/test_redis_integration.py
gandhipratik203: Thanks — addressed in 2b453b6. Added `test_rediss_url_lazy_handshake_reaches_connectivity_layer`. It drives a `tool_pre_invoke` against `rediss://127.0.0.1:1/15` so the lazy TLS handshake actually engages. Default `fail_mode=open` allows the request after logging the backend error; the test then asserts the logged warning is connectivity-shaped (refused / timed out / IO) and explicitly NOT `feature is not enabled` / `InvalidClientConfig` — the latter being the wo-tracker #68217 regression signature.

The fuller "real TLS Redis with self-signed cert + CA fixture" variant is captured as a follow-up in the PR description (last bullet).

plugins/tests/rate_limiter/test_redis_integration.py
gandhipratik203: Thanks — addressed in 2b453b6. Added `test_wipe_only_deletes_keys_under_configured_prefix`. Constructs the plugin with `redis_key_prefix: "custom_pfx"`, seeds both `custom_pfx:user:alice:60` and `other_ns:user:bob:60`, runs `shutdown()` with `mode=disabled`, and asserts `custom_pfx:*` is gone while `other_ns:*` survives. Pins the contract that the wipe respects the configured prefix and never widens to adjacent namespaces.

lucarlig (commented):
Follow-up detailed review pass on the latest head found three remaining P2 findings.

plugins/rust/python-package/rate_limiter/cpex_rate_limiter/rate_limiter.py
```diff
@@ -74,16 +79,237 @@ async def initialize(self) -> None:
...
+            import redis.asyncio as aioredis  # noqa: PLC0415
+        except Exception:
+            return
+        try:
+            client = aioredis.from_url(redis_url, decode_responses=True)
+        except Exception:
+            return
+        try:
+            try:
+                await client.delete(self._wipe_lock_key())
```
lucarlig: [P2] The wipe lock value is always `1`, expires after 30s, and release is an unconditional `delete()`. If a large wipe exceeds the TTL, another worker can acquire the lock, then the stale first worker can delete that newer lock on exit. That breaks the single-flight guarantee under the large/slow-keyspace case it is meant to protect. Use a unique token plus compare-and-delete Lua, or renew ownership while scanning.

plugins/rust/python-package/rate_limiter/cpex_rate_limiter/rate_limiter.py
```diff
@@ -74,16 +79,237 @@ async def initialize(self) -> None:
...
+            import redis.asyncio as aioredis  # noqa: PLC0415
+        except Exception:
+            return False
+        try:
+            client = aioredis.from_url(redis_url, decode_responses=True)
+        except Exception:
+            return False
+        try:
+            try:
+                value = await client.get(f"plugin:{self.config.name}:mode")
```
lucarlig: [P2] The plugin infers admin disable by reading `plugin:<name>:mode` from its configured rate-limit Redis. That couples cleanup to gateway mode storage and assumes the gateway mode key lives in the same Redis DB as counters. If those stores diverge, or the gateway changes persistence, disable can succeed while cleanup silently skips. Prefer an explicit lifecycle/mode-change signal or framework-owned callback.

plugins/tests/rate_limiter/test_redis_integration.py
```diff
@@ -1578,3 +1578,360 @@ async def test_unknown_config_key_emits_warning(self, redis_url_for_integration,
...
+        # — NOT the InvalidClientConfig shape that means rustls itself is
+        # not compiled in. The latter is the failure mode of wo-tracker
+        # #68217 and must never recur.
+        all_messages = " ".join(r.getMessage() for r in caplog.records).lower()
+        assert "tls" not in all_messages or "connect" in all_messages or "refused" in all_messages or "timed out" in all_messages, (
+            "rediss:// failure must surface as a connectivity error, not a "
+            f"TLS feature/config error. Captured logs: "
+            f"{[(r.levelname, r.getMessage()) for r in caplog.records]}"
+        )
+        assert "feature is not enabled" not in all_messages and "invalidclientconfig" not in all_messages, (
```
lucarlig, line 1680: [P2] The `rediss://` lazy-path assertion can pass whenever `tls` is absent from logs, even if no backend call happened or the log is generic. It keeps the negative check for the old feature-disabled failure, but does not prove the path reached network/TLS handshake code. Require a positive network indicator such as connect/refused/timed out/io, or add an optional real TLS Redis fixture.

deny.toml
```diff
@@ -63,6 +63,7 @@ allow = [
     "CC0-1.0",
     "CDDL-1.0",
     "CDDL-1.1",
+    "CDLA-Permissive-2.0",
```
lucarlig: not allowed licence, which package requires it? you might make an exception not add it here

lucarlig (changes requested):

deny.toml
gandhipratik203: Good call — addressed in bdc0740 by dropping the `tls-rustls-webpki-roots` feature from the redis crate. That feature was the only path bringing `webpki-roots` (CDLA-Permissive-2.0) into the dep tree. Without it, the redis crate falls back to `rustls-native-certs` (Apache-2.0/MIT/ISC — already in the workspace allow list) which reads CA certs from the host OS trust store at runtime.

mcp-context-forge ships on UBI 10 Minimal (CA bundle present by default) and AWS ElastiCache uses public-CA chains, so `rustls-native-certs` is operationally equivalent for this deployment with a clean license posture. The `CDLA-Permissive-2.0` entry I added has been removed from `deny.toml` — no exception or allow-list change needed.

plugins/rust/python-package/rate_limiter/cpex_rate_limiter/rate_limiter.py
gandhipratik203: Thanks — split out: PR #74 now only contains the TLS support work, and the wipe code (with the round-1 single-flight fix you'd already flagged) moved to the `feat/rate-limiter-wipe-on-disable-only` branch. This round-2 lock-token + compare-and-delete refinement is being addressed on that branch / follow-up PR, alongside the round-1 wipe-related items. Will link once that PR is open.

plugins/rust/python-package/rate_limiter/cpex_rate_limiter/rate_limiter.py
gandhipratik203: Thanks — same split: wipe code moved to the `feat/rate-limiter-wipe-on-disable-only` branch. The same-Redis-instance assumption you flagged is now documented in that branch's README ("Disabling resets counters" → "Known assumption" paragraph) until the proper framework `on_mode_change` callback lands. Will link the follow-up PR here once opened.

plugins/tests/rate_limiter/test_redis_integration.py
gandhipratik203: Partially addressed in bdc0740. The crypto-provider install now ensures the lazy handshake actually reaches the network (rustls 0.23 was panicking at TLS init without an installed provider — that bug went unobserved under the OR-shortcut assertion).

The assertion shape is now scoped to the wo-tracker #68217 negative signature only ("feature is not enabled" / `InvalidClientConfig` must never appear). A positive "a connect/refused/timed-out indicator was logged" assertion would require the Rust core's `log_exception()` to include the underlying `RedisError` text instead of the generic "error; allowing request" message — currently the error details are dropped at the Python boundary. Tracked as a follow-up alongside the real-TLS-Redis-fixture variant we discussed earlier.

ja8zyjits:
# PR #74 Verification Report

**Date:** 2026-05-02  
**PR:** fix(rate-limiter): production hardening — TLS support for managed Redis  
**Status:** ✅ VERIFIED

## Summary

All concerns outlined in `pr_74.md` are fully addressed by `pr_74.diff`. The implementation matches the current branch (`git diff origin/main` confirms identity).

## Gaps Addressed

### Gap A (HIGH, wo-tracker #68217) - TLS Support
- ✅ redis crate upgraded: 0.27 → 0.32
- ✅ TLS features enabled: `tls-rustls`, `tokio-rustls-comp`
- ✅ Direct rustls dependency with `ring` crypto provider
- ✅ `rediss://` URLs now parse and connect successfully

### Gap B (MEDIUM) - Security Advisory
- ✅ redis 0.32 upgrade removes unmaintained `rustls-pemfile` (RUSTSEC-2025-0134)
- ✅ Dependency tree clean of flagged crates

### Gap C (MEDIUM) - Crypto Provider Panic
- ✅ `ensure_crypto_provider()` function installs ring provider via `OnceLock`
- ✅ Called in `RateLimiterPluginCore::new()` before any Redis client construction
- ✅ Unit test guards against mutation-testing regressions

## Version Bumps

| File | Old | New | Status |
|------|-----|-----|--------|
| `Cargo.toml` | 0.0.5 | 0.0.6 | ✅ |
| `plugin-manifest.yaml` | 0.0.5 | 0.0.6 | ✅ |
| `Cargo.lock` | - | updated | ✅ |

## Test Coverage

### TestRedisTlsSupport (Regression Coverage)
- `test_rediss_url_does_not_fail_with_tls_not_enabled`: Construction with `rediss://` URL succeeds
- `test_rediss_url_lazy_handshake_reaches_connectivity_layer`: Lazy TLS path fails as connectivity, not TLS-feature error

### TestRedisTlsHandshake (End-to-End)
- `redis_tls_url_for_integration`: Module-scoped fixture with self-signed CA + TLS Redis container
- `test_rediss_handshake_succeeds_against_real_tls_redis`: Full rustls handshake verification with counter persistence

## Conclusion

**No gaps found.** Implementation is complete and production-ready.

lucarlig (approved):
LGTM

PR #77, opened 2026-05-03
feat(rate-limiter): wipe Redis counter state on disable transition

### LINKED ISSUE #4576: [FEATURE]: Clear rate-limiter Redis counter state on `disabled` mode transition
### 🧭 Type of Feature

- [x] Enhancement to existing functionality
- [ ] New feature or capability
- [ ] New MCP-compliant server
- [ ] New component or integration
- [ ] Developer tooling or test improvement
- [ ] Packaging, automation and deployment (ex: pypi, docker, quay.io, kubernetes, terraform)
- [ ] Other

### 🧭 Epic

**Title:** Clear rate-limiter persistent counter state when its mode is toggled to `disabled`.

**Goal:** When an operator flips the rate-limiter plugin to `disabled` at runtime, the counter state it has accumulated in Redis should be released, not left behind to silently reapply when the plugin is re-enabled.

**Why now:** This is the concrete follow-up to design question #3 in the parent issue (#4514): _"Should `disabled` mode actually reset Redis state? Today it stops checking counters but doesn't clear them."_ The team has converged on "yes, clear them" as the right behavioural contract for the runtime toggle. This issue tracks delivering that contract.

The user-visible problem today: operators using the runtime toggle as an incident-response release valve (or part of a rollout) expect `disabled` to mean "the limiter's effect is gone." In practice, because the counters persist, the moment the operator flips the plugin back to `enforce`, traffic that was perfectly legitimate during the disabled window can be blocked instantly by counter values inherited from before the toggle. The behaviour is surprising and works against the use case the toggle was added for.

---

### 🧑🏻‍💻 User Story 1

**As an** operator using the runtime mode toggle to disable rate-limiting during an incident or rollout,
**I want** the rate-limiter's persistent counter state to be cleared when I flip the plugin to `disabled`,
**So that** when I subsequently re-enable enforcement, traffic is judged against a fresh window and is not immediately blocked by counters that accumulated before the toggle.

#### ✅ Acceptance Criteria
```gherkin
Scenario: counters cleared on enforce → disabled
  Given the rate-limiter is in `enforce` mode
  And there are non-empty counter keys for the limiter in Redis
  When an operator toggles the plugin's mode to `disabled`
  Then the limiter's counter keys are no longer present in Redis once the toggle has propagated

Scenario: re-enable starts a fresh window
  Given the limiter has just been toggled to `disabled` (and its counters cleared)
  When the limiter is toggled back to `enforce`
  Then a freshly-arriving request under the configured rate is allowed
  And no request is blocked solely on the basis of counter state that existed before the disable
```

---

### 🧑🏻‍💻 User Story 2

**As a** tenant served by a multi-tenant gateway,
**I want** the cleanup that happens on `disabled` to respect tenant scoping and the rate-limiter's key prefix,
**So that** unrelated keys in the same Redis instance (other plugins, shared infra, application data) are unaffected by the toggle.

#### ✅ Acceptance Criteria
```gherkin
Scenario: blast radius is confined to the rate-limiter's keyspace
  Given the rate-limiter has counter keys across multiple tenants
  And there are unrelated keys in the same Redis database that do not belong to the limiter
  When the limiter is toggled to `disabled`
  Then all of the limiter's counter keys (across every tenant) are cleared
  And every unrelated key is left untouched
```

---

### 📐 Design Sketch (optional)

Out of scope here. The agreed direction is "clear on disable"; the approach to clearing — single-flight, batched, prefix-scoped — is being worked through on the investigation branch linked below.

---

### 🔗 MCP Standards Check

- [x] Change adheres to current MCP specifications
- [x] No breaking changes to existing MCP-compliant integrations
- [ ] If deviations exist, please describe them below

The runtime mode toggle is an internal gateway concern; the MCP-visible surface (tool calls, prompts) is unchanged.

---

### 🔄 Alternatives Considered

- **Leave counters in place (status quo).** Operators have to wait for TTL expiry (or manually `FLUSHDB` the appropriate prefix) before re-enable behaves as expected. This is the source of the surprise reported in #4514 design question #3.
- **Clear only on `disabled → enforce` transition.** Same end-state on re-enable, but the counters continue consuming Redis memory during the entire `disabled` window with no functional purpose. Rejected as wasteful and as creating an opportunity for stale state to drift.
- **Operator-driven manual clear** (e.g. an admin endpoint). Useful as a follow-up affordance but doesn't solve the surprise-on-re-enable problem unless operators remember to call it on every toggle.

---

### 📓 Additional Context

- **Parent:** #4514 — specifically design question #3 ("Should `disabled` mode actually reset Redis state?"). This issue is the implementation half of that question once the team converged on "yes."
- **Initial investigation (cpex-plugins):** branch [`feat/rate-limiter-wipe-on-disable-only`](https://github.com/IBM/cpex-plugins/tree/feat/rate-limiter-wipe-on-disable-only) — three commits, including a round of review feedback on single-flight, `UNLINK` vs. `DEL`, and prefix scoping. Treat as a starting point for discussion, not a commitment to ship as-is.
- **Cross-repo tracker:** [IBM/cpex-plugins#49](https://github.com/IBM/cpex-plugins/issues/49) — broader rate-limiter follow-up list.
- **Convergence-test draft PR:** #4511 — relevant because correct toggle behaviour from the operator's perspective depends on both the convergence story (covered in #4514) and the state-cleanup story (this issue).

gandhipratik203:
## Summary

Closes [IBM/mcp-context-forge#4576](https://github.com/IBM/mcp-context-forge/issues/4576) — sub-issue of [#4514](https://github.com/IBM/mcp-context-forge/issues/4514) (rate-limiter runtime mode-toggle convergence).

When an operator flips the rate limiter to `mode: disabled` at runtime, the plugin now clears its accumulated counter state from Redis. Previously the counters survived the toggle, so an immediate re-enable (or a stale-mode worker still serving traffic) would block legitimate requests on counter values inherited from before the disable. After this change, `disabled` truly means "the limiter's effect is gone" — and re-enable starts every user with a fresh window.

Bumps `cpex_rate_limiter` 0.0.4 → 0.0.6.

---

## What changes

When `RateLimiterPlugin.shutdown()` runs, the plugin reads `plugin:<name>:mode` from Redis. If the value is `disabled`, the plugin acquires a distributed single-flight lock (`<prefix>-wipe-lock`, 30s TTL safety net), runs `SCAN + UNLINK` over its configured `redis_key_prefix`, and releases the lock. The wipe is bounded to keys under the plugin's configured prefix; nothing outside that namespace is touched.

The mechanism piggybacks on the framework's existing `reload_tenant` path: when `publish_plugin_mode_change("<name>", "disabled")` lands, every gateway worker's invalidation listener fires `factory.invalidate_all()` → `reload_tenant` per cached context → `manager.shutdown()` → `plugin.shutdown()`. Single-flight via the lock ensures the SCAN+UNLINK runs once across the cluster, even if N workers all process the broadcast simultaneously.

---

## Test coverage

`TestWipeOnDisable` class — 10 tests in `tests/rate_limiter/test_redis_integration.py`:

| Test | Target |
|---|---|
| `test_wipe_fires_when_mode_key_says_disabled` | Happy path |
| `test_no_wipe_when_mode_key_says_enforce` | Graceful pod restart safety |
| `test_no_wipe_when_mode_key_absent` | Fresh-deploy / no-mode-key safety |
| `test_no_wipe_when_redis_unreachable_for_mode_lookup` | Fail-safe on Redis outage |
| `test_wipe_is_idempotent_under_concurrent_shutdown` | 5-instance parallel shutdown converges |
| `test_core_shutdown_runs_even_when_wipe_raises` | Wipe failure doesn't leak Redis connections |
| `test_wipe_skipped_when_another_worker_holds_lock` | Distributed single-flight |
| `test_wipe_only_deletes_keys_under_configured_prefix` | Blast radius confined to configured prefix |
| `test_full_toggle_journey_enforce_disable_reenforce` | Single-instance journey: saturated user is no longer blocked after cycle |
| `test_full_toggle_journey_across_multiple_instances` | N=3 parallel instances: every user gets a fresh window after re-enable, not just the lock-winner's |

<details>

```text
make test-unit                        Rust tests: 58 passed
make test-integration                 pytest: 66 passed
```

</details>

---

## Deployment caveats (worth knowing before merge)

Two pre-conditions for the wipe path to actually fire — intrinsic properties of how the wipe interacts with the framework's plugin lifecycle, worth flagging so operators aren't surprised at start-up:

1. **Plugin must be loaded for the wipe to fire.** When `plugins/config.yaml` has `mode: "disabled"` for the plugin from startup, the framework loader does not instantiate the plugin under any cached manager. No plugin instance ⇒ no shutdown to fire ⇒ no wipe path to run, regardless of how many subscribers receive the broadcast. In practice this means: the wipe only takes effect after the plugin has been in `enforce` mode at least once.
2. **At least one cached manager required.** `factory.invalidate_all()` iterates only *cached* managers; a fresh gateway with zero traffic has zero cached managers. The first request through the gateway warms a manager; subsequent toggle broadcasts then have something to act on.

Neither caveat affects production deployments after the first request. Both are worth documenting on the rate-limiter README so operators understand the start-up timing.

---

## Migration notes

- **No new required config.** Existing deployments work unchanged. The wipe fires automatically on the disable transition.
- **`redis_key_prefix` is the blast-radius boundary.** If multiple rate-limiter instances share one Redis and use the same prefix, a disable on any one wipes counters for all. Default prefix is `"rl"`; operators with multiple limiters should set distinct prefixes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -49,9 +49,9 @@ checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 
 [[package]]
 name = "arc-swap"
-version = "1.9.1"
+version = "1.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6a3a1fd6f75306b68087b831f025c712524bcb19aad54e557b1129cfa0a2b207"
+checksum = "a07d1f37ff60921c83bdfc7407723bdefe89b44b98a9b772f225c8f9d67141a6"
 dependencies = [
  "rustversion",
 ]
@@ -81,9 +81,9 @@ checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
 name = "bitflags"
-version = "2.11.1"
+version = "2.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c4512299f36f043ab09a583e57bceb5a5aab7a73db1805848e8fef3c9e8c78b3"
+checksum = "843867be96c8daad0d758b57df9392b6d8d271134fce549de6ce169ff98a92af"
 
 [[package]]
 name = "block-buffer"
@@ -114,9 +114,9 @@ checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
 
 [[package]]
 name = "cc"
-version = "1.2.61"
+version = "1.2.58"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d16d90359e986641506914ba71350897565610e87ce0ad9e6f28569db3dd5c6d"
+checksum = "e1e928d4b69e3077709075a938a05ffbedfa53a84c8f766efbf8220bb1ff60e1"
 dependencies = [
  "find-msvc-tools",
  "shlex",
@@ -170,9 +170,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.6.1"
+version = "4.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1ddb117e43bbf7dacf0a4190fef4d345b9bad68dfc649cb349e7d17d28428e51"
+checksum = "b193af5b67834b676abd72466a96c1024e6a6ad978a1f484bd90b85c94041351"
 dependencies = [
  "clap_builder",
 ]
@@ -516,9 +516,9 @@ dependencies = [
 
 [[package]]
 name = "hashbrown"
-version = "0.17.0"
+version = "0.16.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4f467dd6dccf739c208452f8014c75c18bb8301b050ad1cfb27153803edb0f51"
+checksum = "841d1cc9bed7f9236f321df977030373f4a4163ae1a7dbfe1a51a2c1a51d9100"
 
 [[package]]
 name = "heck"
@@ -552,13 +552,12 @@ dependencies = [
 
 [[package]]
 name = "icu_collections"
-version = "2.2.0"
+version = "2.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2984d1cd16c883d7935b9e07e44071dca8d917fd52ecc02c04d5fa0b5a3f191c"
+checksum = "4c6b649701667bbe825c3b7e6388cb521c23d88644678e83c0c4d0a621a34b43"
 dependencies = [
  "displaydoc",
  "potential_utf",
- "utf8_iter",
  "yoke",
  "zerofrom",
  "zerovec",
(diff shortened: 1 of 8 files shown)


jonpspri:
Hold in case internal consumer returns with a firm use case.

PR #79, opened 2026-05-05
Automate Rust plugin release tags

### LINKED ISSUE #82: Automate Rust plugin release tags
## Summary
Automate release tag creation for Rust Python package plugins when a plugin version is bumped and merged to main.

## Goal
When a Rust plugin version bump reaches main, CI should run the required checks, create the canonical release tag only after those checks are green, and invoke the publish workflow so PyPI publishing completes without manual tag creation.

## Scope
- Detect plugin Cargo.toml version bumps already surfaced by plugin catalog CI selection.
- Create canonical tags such as rate-limiter-v0.0.5 after required checks pass on main.
- Invoke the reusable release workflow with PyPI publishing enabled after tag creation.
- Update release and testing docs for the automated path.

## Draft PR
- https://github.com/IBM/cpex-plugins/pull/79

lucarlig:
Closes #82

## Summary
- add a post-CI tag creation job for Rust Python package plugin version bumps merged to main
- require build, security, coverage, and documentation jobs to pass before tags are pushed
- call the reusable release workflow after tag creation with PyPI publishing enabled, avoiding reliance on bot tag-push recursion
- cover the workflow gate in catalog tests and make the canonical release-tag test follow the current Cargo version
- update release/testing docs for the automated tag and publish path


msureshkumar88 (changes requested):
The overall approach here is solid — gating on all required CI checks before tagging, calling the release workflow directly instead of relying on bot-tag recursion, and switching the tests to dynamic Cargo.toml version reads rather than hardcoded versions are all the right calls. A few things need addressing before merge, listed by severity.

## Blocking

**`set -e` does not apply inside a piped `while` loop body (silent failure)**

This is the most important one. In bash, `set -euo pipefail` suppresses `set -e` inside the body of a `while read` loop that is on the right-hand side of a pipe. A failed `git push` (network blip, permissions, ref lock) exits non-zero but the loop continues, and the step exits 0. Tags silently fail to push; the job reports green.

The fix is to use process substitution so `set -e` applies inside the loop body, and to discriminate between "already exists" (acceptable) and genuine push failures (fatal):

```bash
while IFS= read -r tag; do
  if [[ ! "${tag}" =~ ^[a-zA-Z0-9._-]+$ ]]; then
    echo "ERROR: unexpected characters in tag '${tag}'; aborting." >&2
    exit 1
  fi
  git tag "${tag}" "${GITHUB_SHA}"
  git push origin "refs/tags/${tag}" || {
    if git ls-remote --exit-code --tags origin "refs/tags/${tag}" >/dev/null 2>&1; then
      echo "Tag ${tag} already exists remotely (concurrent run?); skipping."
    else
      echo "FATAL: push of '${tag}' failed for unknown reason." >&2
      exit 1
    fi
  }
done < <(python3 -c 'import json, os; [print(t) for t in json.loads(os.environ["RELEASE_TAGS"])]')
```

This also collapses the TOCTOU check-then-push into push-then-verify, eliminating the race window between `ls-remote` and `git push` (two concurrent runs could both pass the check and then race to push).

**Tag name validation (shell injection)**

`"${tag}"` prevents word-splitting but not command substitution. A `Cargo.toml` version string containing `$(...)` would execute in the shell. The format-validation guard in the snippet above covers this — a valid semver-based tag only needs `[a-zA-Z0-9._-]`.

## Important

**`fromJson(null)` can crash the `publish-release-tags` matrix at plan time**

GitHub Actions evaluates the `matrix:` expression before the `if:` condition. If `release_validation_tags` is `null` or an empty string, `fromJson` fails at workflow-plan time even for a job that would be skipped. Worth adding a non-empty guard and ensuring `validate-and-detect` always emits at least `'[]'`:

```yaml
if: needs.create-release-tags.result == 'success' && needs.validate-and-detect.outputs.release_validation_tags != '[]'
```

**`secrets: inherit` missing for reusable workflow**

`id-token: write` on the caller job grants the OIDC token to the caller's own steps, but the called reusable workflow needs its own `id-token: write` declaration and any secrets must be passed explicitly. If `release-rust-python-package.yaml` needs any secrets they will silently receive nothing. Worth adding `secrets: inherit` unless you're certain the called workflow is fully self-contained via OIDC.

**`GITHUB_TOKEN` and tag protection rules**

If the repo has tag protection rules matching the release tag pattern, the default `GITHUB_TOKEN` will be rejected. Worth documenting in DEVELOPING.md whether a PAT or GitHub App token is needed, or adding a clear error path.

## Test gaps

Two gate conditions in `create-release-tags` are untested — a future refactor could silently break them:

1. `always()` is not asserted. Without it the job is skipped whenever any `needs` job is skipped (e.g. `mutation-testing` not triggered by a non-Rust change), which is exactly the case this workflow is designed to handle.
2. The `'skipped'` allowance for `mutation-testing` and `release-validation` is not asserted. If someone tightens these to require `'success'` only, tags stop being created for non-Rust pushes with no test catching it.

```python
self.assertIn("always()", create_tags_section)
self.assertIn("needs.mutation-testing.result == 'skipped'", create_tags_section)
self.assertIn("needs.release-validation.result == 'skipped'", create_tags_section)
```

## Minor

- `assertNotIn("mapfile", create_tags_section)` tests a meaningful invariant (bash `mapfile` silently produces an empty array on some shells when input is empty) — worth a one-line comment explaining why, otherwise it reads as a mysterious negative assertion.
- The three tests reading from Cargo.toml dynamically will raise `KeyError`/`FileNotFoundError` instead of a clean `AssertionError` if the file structure changes — low risk given other tests depend on the same files, but worth noting.

lucarlig:
Addressed the requested changes in f4ce16d:

- Reworked the tag creation loop to use process substitution so failures inside the loop are not hidden by a pipeline.
- Added release tag character validation before creating/pushing tags.
- Changed tag publishing to push first, then treat an existing remote tag as the only acceptable push failure.
- Guarded the publish matrix for empty release tag output and made fromJson fall back to [].
- Added secrets: inherit for the reusable release workflow.
- Documented the GITHUB_TOKEN/tag protection requirement in DEVELOPING.md.
- Added workflow assertions for always(), skipped mutation/release-validation allowances, process substitution, tag validation, matrix guard, and secrets inheritance.

Verification: python3 -m unittest tests.test_plugin_catalog passes: 106 tests, 3 skipped.

msureshkumar88 (approved):
All requested changes addressed in f4ce16d (rebased as e3523c1):

- Process substitution loop: no pipeline failure masking.
- Tag character validation before push.
- Push-first, verify-existing-only on failure — clear error path.
- `fromJson(... || '[]')` fallback + `release_validation_tags != '[]'` matrix guard.
- `secrets: inherit` on publish job.
- GITHUB_TOKEN/tag-protection constraint documented in DEVELOPING.md.
- Test coverage added: `always()`, `'skipped'` allowances for `mutation-testing` and `release-validation`, `mapfile` invariant comment.

No new issues introduced. LGTM.

PR #81, opened 2026-05-05
fix: narrow lockfile plugin CI selection

### LINKED ISSUE #80: Narrow CI plugin selection for plugin-scoped lockfile updates
## Problem

PRs that touch one managed plugin plus the root `Cargo.lock` currently expand the Rust Python package CI matrix to every managed plugin. This makes a plugin-scoped dependency or version update much more expensive than intended.

## Expected behavior

When `Cargo.lock` changes together with one or more plugin-specific paths, CI should select only those changed plugins. A lockfile-only or shared workspace change can still conservatively select all plugins.

## Evidence

PR #78 changes only `rate_limiter` paths plus `Cargo.lock`, but CI selected all managed plugins across all OS matrix entries.

## Scope

Update `tools/plugin_catalog.py` CI selection logic and add regression coverage for the lockfile-plus-plugin case.

lucarlig:
## Summary

- keep `Cargo.lock` from forcing all plugins when plugin-specific paths identify the changed plugin set
- preserve conservative all-plugin selection for lockfile-only changes
- add regression coverage for `Cargo.lock` plus one plugin path

Fixes #80

## Validation

- `python3 -m unittest tests.test_plugin_catalog.PluginCatalogTests.test_ci_selection_treats_cargo_lock_change_as_all_plugins tests.test_plugin_catalog.PluginCatalogTests.test_ci_selection_keeps_cargo_lock_with_plugin_change_plugin_scoped`
- `python3 tools/plugin_catalog.py ci-selection . diff a028daee84c0879bda1a38b59111afecbf044b87 c628d106327bf595c024d7fe515abcab7d7ff903`

### DIFF
--- a/tests/test_plugin_catalog.py
+++ b/tests/test_plugin_catalog.py
@@ -1921,6 +1921,57 @@ class PluginCatalogTests(unittest.TestCase):
                 },
             )
 
+    def test_ci_selection_keeps_cargo_lock_with_plugin_change_plugin_scoped(self) -> None:
+        with tempfile.TemporaryDirectory() as tmpdir:
+            root = Path(tmpdir)
+            git = lambda *args: subprocess.run(  # noqa: E731
+                ["git", *args],
+                cwd=root,
+                text=True,
+                capture_output=True,
+                check=True,
+            )
+            git("init")
+            git("config", "user.name", "Test User")
+            git("config", "user.email", "test@example.com")
+            (root / "Cargo.toml").write_text(
+                '[workspace]\nmembers = ["plugins/rust/python-package/rate_limiter", "plugins/rust/python-package/pii_filter"]\n'
+            )
+            rate_limiter = self._create_plugin(root, "rate_limiter")
+            pii_filter = self._create_plugin(root, "pii_filter")
+            (rate_limiter / "src").mkdir()
+            (pii_filter / "src").mkdir()
+            (root / "Cargo.lock").write_text("# seed\n")
+            git("add", ".")
+            git("commit", "--no-verify", "-m", "seed layout")
+            base_sha = git("rev-parse", "HEAD").stdout.strip()
+
+            (rate_limiter / "src" / "lib.rs").write_text("// rate_limiter update\n")
+            (root / "Cargo.lock").write_text("# updated\n")
+            git("add", ".")
+            git("commit", "--no-verify", "-m", "plugin lockfile update")
+
+            result = run_catalog("ci-selection", str(root), "diff", base_sha, "HEAD")
+            self.assertEqual(result.returncode, 0, result.stderr)
+            payload = json.loads(result.stdout)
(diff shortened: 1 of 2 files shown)


msureshkumar88 (changes requested):
**tools/plugin_catalog.py**

L533: ❓ q: `changed_paths` — is this a `list`? If so, `"Cargo.lock" in changed_paths` is O(n) before the loop; the loop itself also iterates it. If callers pass a list, convert to set at top of function to avoid double-scan.

L551: 🟡 risk: `path == "Cargo.lock"` is exact-string match. If `changed_paths` ever contains `./Cargo.lock` or a normalized variant, lockfile is not skipped and falls through to the prefix checks (harmlessly continues, but `has_lockfile_change` stays False → lockfile-only diff returns empty instead of all-plugins). Confirm upstream `git diff --name-only` always produces bare relative paths.

L552–553: ❓ q: `Cargo.lock` removed from global triggers list means a lockfile-only diff now reaches `_changed_plugins_for_records` instead of short-circuiting. Is there any other caller path that previously relied on `Cargo.lock` being in global triggers (e.g., a workflow matrix step that checks the constant directly)?

---

**tests/test_plugin_catalog.py**

L1930: 🔵 nit: test only covers `Cargo.lock` + 1 plugin. Add a case with 2 plugins changed to confirm multi-plugin scoping (avoids silent all-vs-one regression if logic is ever touched).

L1945: 🔵 nit: `README.md` is not a Rust source file — it doesn't exercise the path that would actually cause a lockfile update in practice (source change → dependency bump). Test signal is valid for the path-routing logic, but a `src/lib.rs` touch would better mirror real-world trigger.

---

Overall: logic is correct, fallback preserved, new test covers the intended case. Path-normalization question (L533/551) is the only thing worth a concrete answer before merge.

lucarlig:
Addressed in b18a578.\n\n- Normalized changed paths with leading ./ stripped and materialized a set before the shared-prefix and Cargo.lock checks, so ./Cargo.lock is handled and the Cargo.lock membership check no longer double-scans the original list.\n- Switched the lockfile+plugin test from README.md to src/lib.rs to better match a real source-triggered lockfile update.\n- Extended the test to cover Cargo.lock plus two changed plugins, preserving multi-plugin scoping.\n\nOn the workflow question: .github/workflows/ci-rust-python-package.yaml still has Cargo.lock in both push and pull_request path triggers, so a lockfile-only PR still starts the workflow. The change only removes Cargo.lock from the catalog tool's all-plugin shared-path short-circuit so plugin-scoped changes can stay scoped while lockfile-only diffs still fall back to all plugins.\n\nLocal verification: python3 -m unittest tests/test_plugin_catalog.py passed.

msureshkumar88 (approved):
All five findings from the previous review addressed in b18a578. Path normalization, set conversion, workflow trigger question answered, multi-plugin test case added, and src/lib.rs used as test fixture. LGTM.

PR #83, opened 2026-05-07
Update python, rust, dependencies and version bumping all plugins

brian-hussey:
Updating python dependendencies of all plugin repos in order to update context forge deps too.


lucarlig (approved):
LGTM

PR #99, opened 2026-05-21
chore: refresh plugin dependencies

lucarlig:
## Summary
- update Rust and Python plugin dependencies to the latest stable releases published on or before 2026-05-11
- refresh Cargo and uv lockfiles after applying the 10-day publish cutoff
- bump all affected plugin patch versions and manifests
- adjust source for breaking dependency updates in `rand`, `redis`, and `sha2`

## Validation
- `cargo check --workspace --all-targets`
- `cargo test --workspace`
- `make plugins-validate`
- `git diff --check`
- Rust lock cutoff scan: `count=0`
- Python lock cutoff scan: `count=0`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1403,11 +1403,12 @@ checksum = "63b8176103e19a2643978565ca18b50549f6101881c443590420e4dc998a3c69"
 
 [[package]]
 name = "rate_limiter"
-version = "0.1.2"
+version = "0.1.3"
 dependencies = [
  "cpex_framework_bridge",
  "criterion",
  "log",
+ "mutants",
  "parking_lot",
  "pyo3",
  "pyo3-async-runtimes",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -49,11 +49,28 @@ checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 
 [[package]]
 name = "arc-swap"
-version = "1.9.0"
+version = "1.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a07d1f37ff60921c83bdfc7407723bdefe89b44b98a9b772f225c8f9d67141a6"
+checksum = "6a3a1fd6f75306b68087b831f025c712524bcb19aad54e557b1129cfa0a2b207"
 dependencies = [
  "rustversion",
+]
+
+[[package]]
+name = "arcstr"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "03918c3dbd7701a85c6b9887732e2921175f26c350b4563841d0958c21d57e6d"
+
+[[package]]
+name = "async-lock"
+version = "3.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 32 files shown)


msureshkumar88 (changes requested):
Dep refresh looks solid overall. Three items need attention before merge.

**`rate_limiter/Cargo.toml:2` + `plugin-manifest.yaml`**
`0.1.1 → 0.1.3` skips 0.1.2. Last published tag is `rate-limiter-v0.1.1` — no 0.1.2 tag exists. CLAUDE.md says every change needs a version bump, but these are three commits on one PR. Skipping a version in published history is unusual. Either collapse to a single bump (0.1.2) or add a note explaining why 0.1.2 was intentionally skipped.

**`tools/plugin_catalog.py`**
`serde` added to `encoded_exfil_detection`'s `REQUIRED_PLUGIN_WORKSPACE_DEPENDENCIES`. The `encoded_exfil_detection/Cargo.toml` diff only shows a `base64` patch bump — no `serde` addition. Confirm: is this fixing a pre-existing missing catalog entry, or does `encoded_exfil_detection` actually need `serde` added to its `Cargo.toml` too?

**`rate_limiter/src/redis_backend.rs:308,415,468,529`**
Four functions annotated `#[mutants::skip]` — `evalsha_or_eval`, `eval_fixed`, `eval_sliding`, `eval_token_bucket`. These cover the Redis Lua response parsing paths. The comments explain the reason (needs live Redis harness). Acceptable, but confirm there is a tracking issue or integration test plan so this coverage gap doesn't persist indefinitely.

---

All `redis 1.x` API adaptations look correct (`IoError→Io`, `NoScriptError→Server(NoScript)`, `TypeError→UnexpectedReturnType`, `get_multiplexed_tokio_connection→get_multiplexed_async_connection`). `delay.rs` sampler injection pattern is a clean testability improvement.

lucarlig:
Thanks Suresh. Verified and addressed the review items:

- Collapsed `rate_limiter` from `0.1.3` to `0.1.2` in `Cargo.toml`, `plugin-manifest.yaml`, and `Cargo.lock` so this PR has a single version bump from the last published `0.1.1`. Pushed in `8502190`.
- Rechecked `encoded_exfil_detection`: it does not declare `serde` in the catalog or `Cargo.toml` on the current branch. It only uses `serde_json` plus the base64 patch bump, so no code change needed there.
- For the Redis `#[mutants::skip]` functions: the gap is limited to live Redis Lua/script-cache response paths. Current plan is to cover these with the Redis integration harness rather than mutation-only unit tests, since they need real Redis `EVALSHA`/`NOSCRIPT` behavior.

Validation after the fix:
- `git diff --check`
- `cargo check --workspace --all-targets`
- `make plugins-validate`
- `cargo test --workspace`

msureshkumar88 (approved):
All three requested-changes items resolved:

- **Rate limiter version**: correctly bumped 0.1.1 → 0.1.2 in both `Cargo.toml` and `plugin-manifest.yaml` (commit 8502190f).
- **serde / encoded_exfil_detection catalog**: current `plugin_catalog.py` entry matches the actual `Cargo.toml` — `serde_json` present, no rogue `serde` entry.
- **mutants::skip coverage gap**: `#[mutants::skip]` annotations added to all four Redis eval functions with inline comments documenting the live-harness requirement. `mutants = "0.0.4"` added as a real dependency.

No new issues introduced. ✅

PR #103, opened 2026-05-22
fix: dispatch release workflow after tag creation

### LINKED ISSUE #102: Release validation tags should dispatch publish workflow automatically
## Problem

When a PR with plugin version bumps merges to `main`, `CI Rust Python Package Plugins` creates release tags in `create-release-tags`. Those tags are pushed by workflow automation, so the `Release Rust Python Package Plugin` tag trigger does not reliably start publish runs.

Observed on PR #99: tags were created for the bumped plugins, but PyPI stayed on the previous versions until the release workflow was manually dispatched for each tag.

## Expected

After release tags are created from the merge-to-main workflow, the release workflow should be dispatched automatically for each tag and publish to PyPI.

## Scope

Update the tag creation workflow so merge-to-main release tags trigger the release workflow explicitly after successful tag creation. Keep PR release validation behavior unchanged.

## Acceptance criteria

- Merging a version-bump PR creates tags as today.
- For each created/existing release tag, the workflow dispatches `release-rust-python-package.yaml` with `repository=pypi`.
- Dispatch happens only on `main` push release-tag creation, not during PR validation.
- CI validates the workflow change.

lucarlig:
## Summary

- Fixes #102.
- Dispatches `release-rust-python-package.yaml` explicitly after merge-to-main release tags are created.
- Grants the tag creation job `actions: write` so `gh workflow run` can start the PyPI release workflow.
- Removes the skipped `publish-release-tags` reusable job path and covers the new behavior in the catalog workflow test.

## Verification

- `actionlint .github/workflows/ci-rust-python-package.yaml`
- `git diff --check`

### DIFF
--- a/.github/workflows/ci-rust-python-package.yaml
+++ b/.github/workflows/ci-rust-python-package.yaml
@@ -437,6 +437,8 @@ jobs:
             gh workflow run release-rust-python-package.yaml \
               --ref main \
               -f "tag=${tag}" \
-              -f repository=pypi
-            echo "Dispatched PyPI release workflow for ${tag}."
+              -f repository=pypi \
+              -f publish_enabled=true
+            release_run_url="$(gh run list -w release-rust-python-package.yaml --limit 1 --json url --jq '.[0].url')"
+            echo "Dispatched PyPI release workflow for ${tag}: ${release_run_url}"
           done < <(python3 -c 'import json, os; [print(tag) for tag in json.loads(os.environ["RELEASE_TAGS"])]')
--- a/.github/workflows/ci-rust-python-package.yaml
+++ b/.github/workflows/ci-rust-python-package.yaml
@@ -405,6 +405,7 @@ jobs:
       }}
     runs-on: ubuntu-latest
     permissions:
+      actions: write
       contents: write
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
@@ -413,6 +414,7 @@ jobs:
 
       - name: Create release tags for bumped plugin versions
         env:
+          GH_TOKEN: ${{ github.token }}
           RELEASE_TAGS: ${{ needs.validate-and-detect.outputs.release_validation_tags }}
         run: |
           set -euo pipefail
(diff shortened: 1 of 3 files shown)


msureshkumar88 (changes requested):
## Findings

**`.github/workflows/ci-rust-python-package.yaml`**

`L437-440`: ❓ q: No run ID or URL logged after dispatch. If `gh workflow run` exits 0 but the triggered run fails immediately (bad ref, auth error), there's no breadcrumb linking the tag to its release run. Consider appending `&& gh run list -w release-rust-python-package.yaml --limit 1 --json url --jq '.[0].url'` to surface the URL in the CI log.

`L437`: 🟡 risk: Dispatch relies on `PUBLISH_ENABLED=""` being "not false" in the release workflow's resolve step. The `[[ "${PUBLISH_ENABLED}" == "false" ]]` guard works today, but if that condition ever changes (e.g., to `[[ "${PUBLISH_ENABLED}" != "true" ]]`), dispatches via `workflow_dispatch` silently stop publishing. The `workflow_dispatch` trigger in `release-rust-python-package.yaml` should declare `publish_enabled` as an input with `default: true` to make the intent explicit.

**`tests/test_plugin_catalog.py`**

`L2905-2916`: 🔵 nit: Tests assert `gh workflow run` args appear in `create_tags_section` but don't assert order relative to `git push origin`. Structural guarantee (dispatch only after push) isn't covered. Low-priority since the YAML diff makes it obvious.

---

Approach is sound — `workflow_dispatch` correctly bypasses GitHub's cross-workflow suppression (which blocks `push:tags` from Actions-created tags). The `publish-release-tags` removal + inline dispatch is the right fix for #102. Two items worth addressing before merge: log the dispatched run URL (ops hygiene), and add `publish_enabled: true` as explicit `workflow_dispatch` input in `release-rust-python-package.yaml` (can be a follow-up PR if preferred).

lucarlig:
Addressed the review concerns in 327683b:

- Added explicit `workflow_dispatch.publish_enabled` input with default `true`.
- Passed `publish_enabled=true` from the CI release dispatch.
- Logged the latest release workflow run URL after dispatch.
- Added test coverage for dispatch-after-push ordering, URL logging, and the `workflow_dispatch` default.

Verification: `python3 -m unittest tests.test_plugin_catalog`

msureshkumar88 (approved):
All three findings resolved: run URL logged after dispatch, `publish_enabled` declared as explicit `workflow_dispatch` input with `default: true`, and ordering assertion added to test. LGTM.

PR #107, opened 2026-05-28
Fix payload policy mutation regressions

lucarlig:
Closes 
## Summary

- Fix encoded exfil scanning under CPEX copy-on-write payload isolation by reading mappings through the Python mapping protocol.
- Add policy/isolation regression coverage for mutation-capable plugins that did not already have this issue-specific coverage.
- Bump encoded exfil from 0.3.2 to 0.3.3 and refresh Cargo.lock / secrets baseline.

## Plugin Audit

| Plugin | Mutation-capable for this issue? | Result |
| --- | --- | --- |
| `pii_filter` | Yes | Already covered by PR #89; no extra change in PR #107. |
| `encoded_exfil_detection` | Yes | Fixed mapping scan under CPEX isolation; added policy regressions for `prompt_pre_fetch`, `tool_post_invoke`, `resource_post_fetch`. No `tool_pre_invoke` regression applies because this plugin does not implement or advertise that hook. |
| `secrets_detection` | Yes | Existing implementation already preserved policy-visible mutations; added policy regressions for `prompt_pre_fetch`, `tool_pre_invoke`, `tool_post_invoke`, `resource_post_fetch`. Test-only coverage change, so no plugin version bump. |
| `rate_limiter` | No | Does not return `modified_payload`; no regression needed for this issue. |
| `retry_with_backoff` | No | Does not return `modified_payload`; no regression needed for this issue. |
| `url_reputation` | No | Does not return `modified_payload`; no regression needed for this issue. |

## Root Cause

Encoded exfil directly cast nested containers to `PyDict`. CPEX policy isolation can wrap writable payload fields in copy-on-write mapping objects whose visible values are exposed through the mapping protocol, so direct `PyDict` iteration could miss nested values and produce no policy-visible `modified_payload`.

## Validation

- `CPEX_TEST_PLUGIN_HOOKS=1 uv run pytest ../../../tests/encoded_exfil_detection/test_integration.py -q --asyncio-mode=auto`
- `CPEX_TEST_PLUGIN_HOOKS=1 uv run pytest ../../../tests/secrets_detection -q --asyncio-mode=auto`
- `cargo nextest run -p encoded_exfil_detection`
- `cargo clippy -p encoded_exfil_detection -- -D warnings`
- `cargo fmt --check -p encoded_exfil_detection`

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(^\\.secrets\\.baseline$|package-lock\\.json$|Cargo\\.lock$|uv\\.lock$|go\\.sum$|poetry\\.lock$|Pipfile\\.lock$)",
     "lines": null
   },
-  "generated_at": "2026-05-28T14:43:19Z",
+  "generated_at": "2026-05-29T09:01:43Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -98,7 +98,7 @@
         "hashed_secret": "0962e90797cdf7ff77c8a0a9477a7ca4f493465f",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 895,
+        "line_number": 951,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -106,7 +106,7 @@
         "hashed_secret": "ca3f3092640b07c051f03e52690fda1bcee3f60d",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 916,
+        "line_number": 972,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -114,7 +114,7 @@
         "hashed_secret": "532fdeccb155dce5b528ba039b9a5d201b817e3b",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 935,
+        "line_number": 991,
         "type": "Secret Keyword",
         "verified_result": null
       },
@@ -122,7 +122,7 @@
         "hashed_secret": "cf743b3a58a4d0f91c1d7f5825c0b1b5f7758174",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 972,
+        "line_number": 1028,
         "type": "Base64 High Entropy String",
         "verified_result": null
       },
@@ -130,7 +130,7 @@
         "hashed_secret": "ff1a7ebc241b8eefe9e75e13f20cc22c365ab626",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 972,
+        "line_number": 1028,
         "type": "Secret Keyword",
         "verified_result": null
       }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(^\\.secrets\\.baseline$|package-lock\\.json$|Cargo\\.lock$|uv\\.lock$|go\\.sum$|poetry\\.lock$|Pipfile\\.lock$)",
     "lines": null
   },
-  "generated_at": "2026-05-21T12:19:43Z",
+  "generated_at": "2026-05-28T14:43:19Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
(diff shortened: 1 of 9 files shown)


msureshkumar88 (changes requested):
Three issues need attention before merge.

**`plugins/tests/secrets_detection/test_plugin_hook_results.py`**
`test_tool_post_invoke_redaction_survives_cpex_policy_with_isolated_payload` calls `make_context()` but every other new isolation test in the same file uses `self._context()`. If these two produce different context objects (different plugin config, different mock state), the isolation test is not exercising the same plugin instance as the rest of the class — the isolation guarantee being proven is untestable. Align to `self._context()` (or document why `make_context()` is intentional here).

**`secrets_detection` version not bumped**
CLAUDE.md: "Every change to a core plugin must include a plugin version bump" covering `Cargo.toml`, `plugin-manifest.yaml`, and `Cargo.lock`. The PR description says "Bump encoded exfil from 0.3.2 to 0.3.3" — `secrets_detection` is not mentioned. If the new isolation tests land inside `plugins/rust/python-package/secrets_detection/tests/` (plugin package directory), `secrets_detection/Cargo.toml` and `plugin-manifest.yaml` need a patch bump too.

**`encoded_exfil_detection` missing `tool_pre_invoke` isolation regression**
The plugin audit table lists `secrets_detection` as covering all four hooks (`prompt_pre_fetch`, `tool_pre_invoke`, `tool_post_invoke`, `resource_post_fetch`), but `encoded_exfil_detection` only gets three (`prompt_pre_fetch`, `tool_post_invoke`, `resource_post_fetch`) — `tool_pre_invoke` is absent. If `encoded_exfil_detection` implements `tool_pre_invoke` with a mutating code path, this is a coverage gap for the exact regression being fixed. Confirm: does `encoded_exfil_detection` implement `tool_pre_invoke`? If yes, add the isolation test; if no, add a note in the PR description explaining why.

lucarlig:
@msureshkumar88 thanks, checked all three points:

- `test_tool_post_invoke_redaction_survives_cpex_policy_with_isolated_payload` is consistent with the surrounding `secrets_detection` tests in this branch: this test module uses `make_context()` throughout and does not define a local `self._context()` helper. No code change needed there.
- I did not bump `secrets_detection`: the PR only adds repo-level regression tests under `plugins/tests/secrets_detection/...`; it does not change the plugin package itself. A version bump would not mean anything for test-only coverage. I updated the PR body to make that explicit.
- `encoded_exfil_detection` does not implement or advertise `tool_pre_invoke`. Its manifest and plugin class list only `prompt_pre_fetch`, `tool_post_invoke`, and `resource_post_fetch`, so I updated the PR body with that note instead of adding a non-applicable test.

msureshkumar88 (commented):
## Review: Fix payload policy mutation regressions

### Summary
Targeted fix for a real production gap: `scan_container` was casting to `PyDict` directly, which silently misses CoW mapping wrappers that CPEX policy isolation can inject. Switching to `PyMapping` + `items()` is the correct protocol-based approach.

### Core Fix — `src/lib.rs`

```rust
// Before
if let Ok(dict) = container.cast::<PyDict>() {
    for (key, value) in dict.iter() { ... }

// After
if let Ok(mapping) = container.cast::<PyMapping>() {
    for item in mapping.items()?.iter() {
        let item = item.cast::<PyTuple>()?;
        let key = item.get_item(0)?;
        let value = item.get_item(1)?;
```

**Correct.** `PyMapping` covers any object implementing `__getitem__`/`keys()`/`items()`, which includes both plain `dict` and CPEX's CoW wrappers. `dict.iter()` in PyO3 only works on actual `PyDict` — casting fails silently on wrapper types, meaning `scan_container` returned an empty result set.

One note: `mapping.items()` materializes a Python list before iteration (vs the zero-copy `dict.iter()`). For typical payload sizes this is negligible, but worth noting if this path ever hits very large nested objects.

### Test Coverage

Policy regression pattern is consistent and correct across all 8 new tests:
1. `wrap_payload_for_isolation` → apply plugin → `apply_policy` → assert original untouched + filtered has redaction.

The assertions that verify **both** the original payload is immutable **and** the filtered result contains the redaction are exactly right.

**Encoded exfil** — 3 hooks covered (`prompt_pre_fetch`, `tool_post_invoke`, `resource_post_fetch`). No `tool_pre_invoke` regression, matching the plugin's advertised hook set. Correct omission.

**Secrets detection** — 4 hooks covered. Test-only change, no version bump. Correct.

### Missing: Rust unit test for `scan_container` with CoW-like input

The integration tests exercise the full Python path, but there's no inline `mod tests` entry in `lib.rs` that unit-tests `scan_container` directly against a mock mapping object. If someone later refactors the cast logic, the Rust-level regression would only surface in Python integration tests, which are slower and harder to bisect. A small Rust unit test with a custom `PyMapping`-compatible object would close this gap.

Not a blocker — the integration tests do validate the fix end-to-end.

### Version Bump

- `encoded_exfil_detection`: 0.3.2 → 0.3.3 in `Cargo.toml`, `plugin-manifest.yaml`, `Cargo.lock`. Correct.
- `secrets_detection`: no bump. Test-only change. Correct per project conventions.

### `.secrets.baseline`

Field reordering (`is_secret` before `is_verified`) is a tool-regenerated cosmetic change. Line number shifts are expected from new test content. No concerns.

### Verdict

**Approve with one suggestion.** The core fix is minimal and correct. Test coverage for the policy isolation scenario is thorough. The only gap is the missing Rust-level unit test for `scan_container` — worth a follow-up issue if not addressed here.

msureshkumar88:
All three prior points acknowledged — responses are accurate and complete:

- `make_context()` vs `self._context()`: confirmed consistent with the secrets_detection test module pattern. No issue.
- No `secrets_detection` version bump: tests land in `plugins/tests/secrets_detection/` (repo-level, outside the plugin package directory), so CLAUDE.md's version bump requirement does not apply. Correct call.
- No `tool_pre_invoke` for `encoded_exfil_detection`: plugin does not implement or advertise that hook. Omission is correct, not a gap.

---

Additional observations for the record:

**Name inconsistency (minor)**
Two of the three new encoded_exfil isolation tests end in `_with_isolated_payload`:
- `test_prompt_pre_fetch_redaction_survives_cpex_policy_with_isolated_payload`
- `test_tool_post_invoke_redaction_survives_cpex_policy_with_isolated_payload`

The third does not:
- `test_resource_post_fetch_redaction_survives_cpex_policy` ← missing suffix

Same pattern in secrets_detection: `test_resource_post_fetch_redaction_survives_cpex_policy` vs `test_prompt_pre_fetch_redaction_survives_cpex_policy_with_isolated_payload`. Not a functional issue, but inconsistent naming makes grep/discovery harder. Worth aligning in a follow-up.

**`mapping.items()` materializes before iteration**
`dict.iter()` in PyO3 is incremental; `mapping.items()?` builds a full Python list of 2-tuples first. For typical CPEX payload sizes this is negligible — flagging only so it's a known trade-off, not a surprise if profiling ever surfaces it.

**`cast::<PyMapping>()` is forward-compatible**
Worth noting as a positive: the fix is more defensive than strictly required for today's CoW wrappers. Any future CPEX isolation layer that implements the Python mapping protocol will work without another `scan_container` change.

---

No blockers. Ready to approve.

lucarlig:
@msureshkumar88 addressed the latest follow-up observations in `c7eb105`:

- Renamed the two resource isolation tests to include `_with_isolated_payload` for grep/discovery consistency.
- Added a Rust unit regression for `scan_container` using a non-dict `collections.abc.Mapping` wrapper so the PyMapping path is covered below the Python integration layer.

msureshkumar88 (approved):
All requested changes addressed. Verified in c7eb105:

- `make_context()` usage: consistent with the secrets_detection test module pattern (no local `self._context()` helper). No issue.
- No `secrets_detection` version bump: tests land in `plugins/tests/secrets_detection/` (repo-level, outside the plugin package), so CLAUDE.md version bump requirement does not apply. Correct.
- No `tool_pre_invoke` test for `encoded_exfil_detection`: plugin does not implement that hook. PR body updated. Correct omission.
- Naming consistency: `test_resource_post_fetch_*` tests in both plugins now carry the `_with_isolated_payload` suffix. Fixed.
- Rust unit regression: `test_scan_container_detects_mapping_wrapper_values` added at `lib.rs:872` — exercises `scan_container` against a non-dict `collections.abc.Mapping` wrapper directly at the Rust level. Gap closed.

Approved.

PR #108, opened 2026-05-29
Fix padded base64 secret detection

lucarlig:
## Summary
- Fix padded base64 secret detection so encoded secrets with trailing padding are matched correctly
- Avoid consuming trailing separators so adjacent padded base64 secrets are detected independently
- Add scanner coverage for double-equals padding, single-equals padding, adjacent padded values, plus/slash payload characters, and base64-char boundary rejection
- Bump the secrets-detection plugin version metadata

## Tracker
- https://github.ibm.com/WatsonOrchestrate/wo-tracker/issues/70713

## Testing
- make test-unit
- make fmt-check clippy test-integration
- cargo mutants -p secrets_detection --in-diff cargo-mutants.diff

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1718,7 +1718,7 @@ checksum = "94143f37725109f92c262ed2cf5e59bce7498c01bcc1502d7b9afe439a4e9f49"
 
 [[package]]
 name = "secrets_detection"
-version = "0.3.3"
+version = "0.3.4"
 dependencies = [
  "cpex_framework_bridge",
  "criterion",
--- a/plugins/rust/python-package/secrets_detection/Cargo.toml
+++ b/plugins/rust/python-package/secrets_detection/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "secrets_detection"
-version = "0.3.3"
+version = "0.3.4"
 edition.workspace = true
 authors.workspace = true
 license.workspace = true
--- a/plugins/rust/python-package/secrets_detection/cpex_secrets_detection/plugin-manifest.yaml
+++ b/plugins/rust/python-package/secrets_detection/cpex_secrets_detection/plugin-manifest.yaml
@@ -1,6 +1,6 @@
 description: "Detect likely credentials and secrets in prompt input, tool output, and resource content"
 author: "ContextForge Contributors"
-version: "0.3.3"
+version: "0.3.4"
 kind: "cpex_secrets_detection.secrets_detection.SecretsDetectionPlugin"
 available_hooks:
   - "prompt_pre_fetch"
--- a/plugins/rust/python-package/secrets_detection/src/patterns.rs
+++ b/plugins/rust/python-package/secrets_detection/src/patterns.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use regex::Regex;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::sync::LazyLock;
 
 pub static PATTERNS: LazyLock<HashMap<&'static str, Regex>> = LazyLock::new(|| {
@@ -60,13 +60,18 @@ pub static PATTERNS: LazyLock<HashMap<&'static str, Regex>> = LazyLock::new(|| {
     );
     patterns.insert(
         "base64_24",
+        // Minimum valid padded base64 secrets are 24 total chars: 22 data
+        // chars plus `==`, or 23 data chars plus `=`.
         Regex::new(
-            r"(?:^|[^A-Za-z0-9+/])((?:[A-Za-z0-9+/]{24,}={0,2}|[A-Za-z0-9+/]{22}==|[A-Za-z0-9+/]{23}=))(?:$|[^A-Za-z0-9+/=])",
+            r"(?:^|[^A-Za-z0-9+/])((?:[A-Za-z0-9+/]{24,}={0,2}|[A-Za-z0-9+/]{22}==|[A-Za-z0-9+/]{23}=))",
         )
-            .expect("valid base64_24 regex"),
+        .expect("valid base64_24 regex"),
     );
     patterns
 });
+
+pub static CAPTURE_PATTERNS: LazyLock<HashSet<&'static str>> =
+    LazyLock::new(|| HashSet::from(["base64_24"]));
 
 #[cfg(test)]
 mod tests {
--- a/plugins/rust/python-package/secrets_detection/src/patterns.rs
+++ b/plugins/rust/python-package/secrets_detection/src/patterns.rs
@@ -60,7 +60,10 @@ pub static PATTERNS: LazyLock<HashMap<&'static str, Regex>> = LazyLock::new(|| {
     );
     patterns.insert(
         "base64_24",
-        Regex::new(r"\b[A-Za-z0-9+/]{24,}={0,2}\b").expect("valid base64_24 regex"),
+        Regex::new(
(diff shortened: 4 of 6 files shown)


msureshkumar88 (changes requested):
## Review: Fix padded base64 secret detection

Fix is directionally correct — old `\b` failed on `+`/`/` chars. New consuming-boundary approach works but has a genuine edge-case bug and design debt. **Tests were explicitly not run** (`Testing: Not run in this branch`) — this alone should block merge for a security-sensitive plugin.

---

### 🔴 Bugs

**`src/scanner/text_scan.rs` L28-40 + `src/patterns.rs` L63 — consuming boundary drops adjacent secret:**
The trailing `(?:$|[^A-Za-z0-9+/=])` *consumes* the separator character. `captures_iter` finds non-overlapping matches, so when two base64 secrets share a single-char separator (e.g. `token1==,token2==`), the `,` is consumed by the first match's trailing group. The next match position starts at `t` of `token2`, where the leading `(?:^|[^A-Za-z0-9+/])` finds no non-base64 boundary char — second secret silently dropped. Fix: use lookahead/lookbehind via `fancy-regex` crate, or post-filter with `find_iter` and strip boundary chars from the captured span.

**`src/scanner.rs` — `{23}=` single-padding branch untested:**
Only `==` double-padding is tested. The regex has a distinct `[A-Za-z0-9+/]{23}=` branch — zero test coverage for it.

---

### 🟡 Risks

**`src/scanner/text_scan.rs` L28 — `if *name == "base64_24"` dispatch in hot path:**
String comparison per pattern per call couples the generic loop to one pattern's implementation detail. Every future capture-group pattern adds another `else if` branch. Fix: store a `uses_captures: bool` flag in the `PATTERNS` map entry (or a separate `CAPTURE_PATTERNS` set) and dispatch generically.

**Tests not run — security-sensitive change:**
PR body explicitly states `Testing: Not run in this branch`. This is a secrets-detection plugin — false negatives mean real credentials leak undetected. Must be blocked until `make test-all` and `make test-integration` pass in-branch.

**`src/scanner.rs` — no test for two adjacent base64 secrets:**
Without it, the consuming-boundary bug above is undetectable by the test suite.

**`src/scanner.rs` — no test with `+`/`/` inside the secret:**
The old `\b` regex failed specifically on `+` and `/` (non-word chars break word boundary). The new test uses only alphanumeric chars (`mZ8qL2vYwT1pNc4Rb6HxUg==`). Add a test with `+` or `/` in the payload to regression-lock the original bug fix.

---

### 🔵 Nits

**`src/scanner.rs` L329-331:** `["mZ8qL2vYwT1p", "Nc4Rb6HxUg", "=="].concat()` — three-part array concat obscures the test payload. Use `"mZ8qL2vYwT1pNc4Rb6HxUg=="` directly.

**`src/patterns.rs` L64-66:** `.expect("valid base64_24 regex")` has extra leading indentation vs. call-chain style used elsewhere in the file.

---

### Documentation gaps

**`src/scanner/text_scan.rs` L28:** No comment explaining why `base64_24` diverges from the generic `find_iter` path. Add: `// base64_24 uses a capture group to exclude boundary chars; extract group 1 (Rust regex has no lookbehind)`

**`src/patterns.rs` L63:** No comment explaining why `{22}==` and `{23}=` alternatives are separate from `{24,}={0,2}`. The reason (minimum valid padded base64 is 24 total chars: 22 data + `==` or 23 data + `=`) is non-obvious to future readers.

---

### ❓ Questions

**No linked issue:** PR has no `Closes #XX` or `Fixes #XX`. What issue does this trace to?

**URL-safe base64:** New regex still only covers standard base64 (`+`/`/`). URL-safe base64 uses `-`/`_`. Intentional exclusion or gap?

---

### Summary

| Category | Finding | Severity |
|---|---|---|
| Bug | Consuming boundary drops second adjacent secret | 🔴 |
| Bug | `{23}=` single-padding branch untested | 🔴 |
| Risk | Tests not run per author's own statement | 🟡 |
| Risk | Pattern-name string dispatch in hot path, will accumulate | 🟡 |
| Risk | No test with `+`/`/` chars — regression for original bug | 🟡 |
| Risk | No test for two adjacent secrets | 🟡 |
| Nit | Test payload split across array concat | 🔵 |
| Nit | `.expect` indentation | 🔵 |
| Doc | Missing comment on `captures_iter` divergence | gap |
| Doc | Missing comment on alternation rationale | gap |
| Issue | No linked issue number | gap |
| Question | URL-safe base64 (`-`/`_`) — intentional? | ❓ |
| Breaking | Not breaking — detection-only increase | ✅ |

lucarlig:
Addressed the review feedback in 9b5ca96:

- Removed the consuming trailing boundary from base64_24; the scanner now post-checks the next char so adjacent secrets separated by one char are both detected.
- Replaced the hard-coded base64_24 dispatch with a generic capture-pattern set.
- Added coverage for adjacent padded base64 secrets, single-equals padding, and plus/slash payload characters.
- Simplified the split test literal and added comments explaining the capture path and padded-length alternatives.

Validation passed:
- make test-unit: 68 passed
- make fmt-check clippy test-integration: fmt/clippy passed, 83 integration tests passed

URL-safe base64 remains out of scope for this change; current behavior continues to target standard base64.

lucarlig:
CI mutation failure addressed in 7b51d1a.

Root cause: the tests did not prove that a padded base64 candidate followed by another base64 character is rejected. That left two boundary-check mutants alive.

Added coverage for that case. Local validation now passes:
- make test-unit: 69 passed
- make fmt-check clippy test-integration: fmt/clippy passed, 83 integration tests passed
- cargo mutants -p secrets_detection --in-diff cargo-mutants.diff: 7 mutants tested, 5 caught, 2 unviable, 0 missed

Also added tracker reference: https://github.ibm.com/WatsonOrchestrate/wo-tracker/issues/70713

msureshkumar88 (approved):
## Re-review: All requested changes addressed

All 9 findings from the previous review are resolved.

### ✅ Verified fixed

| Finding | Resolution |
|---|---|
| 🔴 Consuming boundary drops adjacent secret | Trailing boundary removed from regex. Scanner extracts group 1 and post-checks `is_base64_boundary_char(char_after_match)` at `text_scan.rs:35-37` — clean separation of concerns |
| 🔴 `{23}=` single-padding untested | `detects_single_padding_base64_secret` covers the `[A-Za-z0-9+/]{23}=` branch |
| 🟡 `if *name == "base64_24"` string dispatch in hot path | Replaced by `CAPTURE_PATTERNS.contains(name)` — pattern dispatch is now data-driven and extensible |
| 🟡 Tests not run | PR body now lists `make test-unit`, `make fmt-check clippy test-integration`, `cargo mutants` |
| 🟡 No adjacent-secrets test | `detects_each_adjacent_padded_base64_secret` verifies both `{22}==` and `{23}=` adjacent |
| 🟡 No `+`/`/` payload test | `detects_padded_base64_secret_with_plus_and_slash` locks the original `\b` regression |
| 🔵 Three-part `concat()` | `let sample = "mZ8qL2vYwT1pNc4Rb6HxUg=="` — direct string |
| 🔵 `.expect()` indentation | Fixed — consistent with call-chain style |
| Doc gaps | Both comments added: capture-group divergence in `text_scan.rs` and alternation rationale in `patterns.rs` |

### 🔵 New nit (non-blocking)

`text_scan.rs:134` — `is_base64_boundary_char` returns `true` when the adjacent char **is** a base64 char (signalling the match should be rejected). The name reads backwards — `is_base64_char` or `is_base64_alphabet_char` would remove the double-negative at the call site (`if is_base64_boundary_char(...) { continue }`). Ignore or rename at your discretion.

---

Fix design is correct. The two-part approach (no trailing boundary in regex + trailing char post-check in scanner) neatly solves the adjacent-detection problem without `fancy-regex`. Version bump and all three metadata files updated consistently.

PR #110, opened 2026-06-15
fix(rate-limiter): enable rustls default features so TLS 1.2 Redis connects

gandhipratik203:
## Summary

cpex-rate-limiter pinned rustls with `default-features = false` and only `["ring", "std"]`,
so rustls compiled without TLS 1.2 (1.3-only); the redis crate's `tls-rustls` feature does
not enable `tls12` either. A Redis endpoint that offers only TLS 1.2 (e.g. AWS ElastiCache)
then shares no protocol version with the client: the handshake fails ("unexpected end of
file") and the plugin returns BACKEND_UNAVAILABLE, even though redis-py (OpenSSL) connects
to the same endpoint.

## Fix

Enable rustls's default features (which include `tls12`) and keep the `ring` provider,
matching the pinning used by contextforge-gateway-rs. Restores TLS 1.2 while keeping 1.3.
Bumps the package to 0.1.4 and adds a regression test that fails to build if `tls12` is
dropped.

## Validation

Confirmed against a live AWS ElastiCache instance: the stock build returns
BACKEND_UNAVAILABLE; this build connects and writes rate-limit keys. Release-validation
wheels build on all targets (linux x86_64/aarch64/ppc64le/s390x, windows, macos).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1412,7 +1412,7 @@ checksum = "63b8176103e19a2643978565ca18b50549f6101881c443590420e4dc998a3c69"
 
 [[package]]
 name = "rate_limiter"
-version = "0.1.3"
+version = "0.1.4"
 dependencies = [
  "cpex_framework_bridge",
  "criterion",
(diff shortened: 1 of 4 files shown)


lucarlig (approved):
LGTM

PR #122, opened 2026-06-18
docs: refresh plugin readmes

lucarlig:
## Summary
- update the root plugin catalog to list all managed plugins
- refresh stale secrets detection and URL reputation READMEs to match current behavior and README shape
- update plugin README runtime dependency text to match the current package dependency

## Tests
- make plugins-validate


gandhipratik203 (approved):
LGTM!

PR #127, opened 2026-06-24
Make release tag creation rerunnable

lucarlig:
## Summary
- skip tag creation when the release tag already exists remotely at the current commit
- keep failing hard if an existing local or remote tag points at a different commit
- continue dispatching the release workflow after an existing-tag skip so partial release reruns can recover

## Verification
- bash syntax check for the workflow run block
- YAML parse check for `.github/workflows/ci-rust-python-package.yaml`
- `git diff --check -- .github/workflows/ci-rust-python-package.yaml`
- fake-remote rerun test with one existing tag and one missing tag

## Incident notes
The failed run had `encoded-exfil-detection-v0.3.5` already created at `a925837ddcca879467b81f0001464ac58609d442`; the other expected tags were still missing. The old loop failed at local `git tag` before reaching its remote-exists recovery path.

### DIFF
--- a/.github/workflows/ci-rust-python-package.yaml
+++ b/.github/workflows/ci-rust-python-package.yaml
@@ -429,15 +429,32 @@ jobs:
               echo "ERROR: unexpected characters in tag '${tag}'; aborting." >&2
               exit 1
             fi
-            git tag "${tag}" "${GITHUB_SHA}"
-            git push origin "refs/tags/${tag}" || {
-              if git ls-remote --exit-code --tags origin "refs/tags/${tag}" >/dev/null 2>&1; then
-                echo "Tag ${tag} already exists remotely; skipping."
-              else
-                echo "FATAL: push of '${tag}' failed." >&2
+            if remote_sha="$(git ls-remote --exit-code --tags origin "refs/tags/${tag}" | awk '{print $1}')"; then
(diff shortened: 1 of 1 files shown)


msureshkumar88 (approved):
Verified locally via sandboxed git sim (bare \"origin\" + lightweight tags) replaying the incident scenario:
- existing-tag-v1.0.0 already at HEAD sha → skip tag creation, still dispatches release workflow (recovers partial rerun) ✅
- missing-tag-v1.0.0 → created + pushed ✅
- tag exists at a *different* sha → FATAL, exit 1, no dispatch ✅ (fail-hard preserved per requirement)

Also ran YAML parse + bash syntax check on the modified workflow file — clean.

Nits (non-blocking):
- L433: `awk '{print $1}'` takes the first `ls-remote --tags` line. Fine since this script only ever creates lightweight tags (`git tag "${tag}" "${GITHUB_SHA}"`, no `-a`), but if a release tag is ever annotated manually, the peeled `refs/tags/<tag>^{}` line could shift which sha lands in `$1`.
- L443-451: `remote_sha` re-derived with the same `ls-remote | awk` pattern in both the initial check and the push-failure recovery block. Duplication is reasonable given `run:` blocks don't share functions across steps.

LGTM, approving.

PR #128, opened 2026-07-01
feat(pii_filter): OpenTelemetry trace context + gated metrics on result.metadata

### LINKED ISSUE #27: Track OpenTelemetry support at the Rust plugin boundary
## Summary
This tracks the `cpex-plugins` portion of the OpenTelemetry/metrics work that was originally bundled into [IBM/mcp-context-forge PR #4069](https://github.com/IBM/mcp-context-forge/pull/4069).

That PR mixed changes across multiple repositories. This issue keeps the plugin-side work scoped to `IBM/cpex-plugins` so it can be implemented and reviewed independently.

## Implementation Direction
This does **not** need to be blocked on a new framework contract. CPEX already provides an extension framework that can carry this data through plugin hooks.

Code checked in [`contextforge-org/cpex`](https://github.com/contextforge-org/cpex):
- `cpex.framework.extensions.request.RequestExtension` already includes `trace_id` and `span_id` for distributed tracing.
- `cpex.framework.extensions.extensions.Extensions` carries typed request/context metadata plus a mutable `custom` slot.
- `cpex.framework.base.HookRef` accepts hook signatures with `payload`, `context`, and optional `extensions`.
- `cpex.framework.manager.PluginExecutor` passes filtered extensions to hooks that accept them.
- `cpex.framework.models.PluginResult` supports both `modified_extensions` and `metadata`.

Use the CPEX extension framework rather than inventing a separate plugin-side telemetry transport. [IBM/mcp-context-forge PR #3754](https://github.com/IBM/mcp-context-forge/pull/3754) and [contextforge-org/cpex issue #43](https://github.com/contextforge-org/cpex/issues/43) can still inform rollout details, but they should not be treated as hard blockers for plugin-side work.

## Scope for this repo
- accept and propagate trace IDs through the Rust plugin boundary using the CPEX extension framework
- pass the framework-provided `Extensions` object through the Python/PyO3 packaging layer where applicable
- return plugin-side metrics through `PluginResult.metadata`, `Extensions.custom`, or a dedicated CPEX metrics extension if one exists by implementation time
- keep each plugin self-contained and avoid assuming an external OpenTelemetry collector/server or similar infrastructure is present
- cover plugin-side tracing and metrics behavior with focused Rust and Python integration tests where applicable

## Out of Scope
- gateway/runtime changes that belong in `IBM/mcp-context-forge`
- defining a separate telemetry transport in this repository
- requiring an external OpenTelemetry server, collector, or similar service for plugin functionality
- cross-repo rollout coordination beyond linking companion issues

## Background
- related to the work originally proposed in [IBM/mcp-context-forge PR #4069](https://github.com/IBM/mcp-context-forge/pull/4069)
- CPEX extension framework reference: [`contextforge-org/cpex`](https://github.com/contextforge-org/cpex)
- created after the earlier cross-repo PR was closed so the work can be tracked per repository instead of as one cross-repo change

msureshkumar88:
## Summary

Closes #27. Wires the `pii_filter` Rust plugin into OpenTelemetry distributed tracing: accepts trace context in, returns gated observability metrics out — using the **metadata-channel**.

## Design decision: `PluginResult.metadata` vs `Extensions.custom`

Issue #27 sanctions two channels for metrics-out. This PR uses `result.metadata`, not `Extensions.custom`, because (verified in the CPEX framework source, not assumed):

- **The executor already accumulates `result.metadata` across the whole hook chain** (`cpex/framework/manager.py`) into a single merged dict — no plugin-side merge helper needed.
- **`Extensions.custom` is multi-plugin lossy by construction**: the executor propagates it via wholesale replace ("last writer wins"), so a second plugin emitting on the same hook silently clobbers the first plugin's metrics unless a fold-at-end-of-chain workaround is added on top. `metadata` has no such loss since each plugin's key is namespaced.
- **The gateway's existing read-path already consumes `result.metadata`** as a flat dict elsewhere in the codebase; the `modified_extensions` path has no existing reader.
- **The reserved-key surface is narrow** (`_decision_plugin` only), so a namespace key like `"pii_filter"` can't collide with framework internals.

Trade-off accepted: `metadata` is an untyped dict with no single-writer chokepoint — mitigated here by per-plugin namespacing (S2) and tests, not by framework-level tier governance (which, in this cpex build, has zero runtime call sites regardless of channel).

## What changed

- All 4 hooks (`prompt_pre_fetch`, `prompt_post_fetch`, `tool_pre_invoke`, `tool_post_invoke`) accept an optional `extensions` parameter carrying trace context, backward-compatible with legacy 2-arg callers (`extensions` defaults to `None`).
- When `extensions.request.trace_id` is present, the plugin returns namespaced, **allow-listed** metrics on `result.metadata["pii_filter"]`: `total_detections`, `total_masked`, `detection_types` (≤32, sorted, deduped), `stage`. Counts and type-names only — **never the matched PII value** (S1). No trace_id is present → no metrics are emitted at all (metrics are opt-in-by-tracing, not always-on).
- Legacy `context.metadata["pii_filter_stats"]` / `["pii_detections"]` writes removed — no double-write, one authoritative channel (P-4). See the README Migration Note for the 0.3.6 breaking-change guidance for anyone reading the old keys.
- Version bumped 0.3.5 → 0.3.6; `crates/framework_bridge` bumped 0.1.1 → 0.1.2 for the new `build_framework_object_dyn` helper (variable-length kwargs, needed because the number of result fields varies by branch).
- Metric-schema contract documented at `docs/superpowers/contracts/pii_filter-metrics-schema.md` — the explicit allow-list/deny-list and the cross-repo source of truth the gateway's consumer (S4 sanitizer) validates against.

## CI stabilization (follow-up commits, after initial push)

Three CI checks were failing on push and got fixed in `e7036a0`, all verified green before this description was written:

- **`mutation-testing`**: cargo-mutants found 2 surviving mutants in the new `handle_nested_stage` code (a deleted `!` on the masked-detection branch, and the `push_metrics_kwarg` no-op mutant) — neither was exercised by a Rust-level test (the Python integration tests that *did* cover this behavior don't count for `cargo mutants`, which only runs `cargo test`). Added a Rust unit test that installs a fake `cpex.framework` module and drives `prompt_pre_fetch` end-to-end, asserting both the masked-detection log line and the metrics kwarg. Confirmed both mutants now caught via a local `cargo mutants --in-diff` run matching CI's exact scope.
- **`test`**: unrelated to this feature — `deny.toml`'s RUSTSEC-2026-0176/0177 advisory-ignore entries (both tied to the pre-existing `pyo3-stub-gen`/PyO3 0.29 incompatibility) had a hardcoded `expires 2026-06-29` and lapsed while this branch was open. Bumped to `2026-08-01`; this would have failed on `main` too, independent of this PR.
- **`detect-secrets`**: 2 false-positive "Secret Keyword" findings — both are the same test-fixture line (`secret = "alice@example.com"`) in the S1 leakage test and the plan doc, flagged only because the variable is literally named `secret`. Audited as not-a-secret in `.secrets.baseline`.

## Test plan

- [x] Rust unit tests (`cargo test --lib`) — 75 passing, including the emission gate, S1 no-content-leak test, S3 bounds test (32-item cap, sorted, deduped), and the new masked-detection/metrics-kwarg mutation-killer test
- [x] Python integration tests (`make test-integration`) — 26 passing, including forwarding/back-compat and the S1 leakage test (metrics + logs)
- [x] `cargo mutants -p pii_filter --in-diff` — 0 missed (was 2)
- [x] `cargo clippy -p pii_filter --all-targets -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] `make verify-stubs` — passes (live `make stub-gen` regeneration is blocked by a pre-existing, repo-wide `pyo3-stub-gen`/`pyo3` incompatibility that predates this branch; hardcoded stub text updated manually)
- [x] `detect-secrets`, `test` (cargo-deny advisory freshness), `mutation-testing` — all green as of `e7036a0`

## Companion PR & merge order

**Merge this PR first, then tag/release before merging the gateway PR.**

Gateway-side work (`IBM/mcp-context-forge` PR #5470) consumes this plugin's published behavior via the `cpex-pii-filter` PyPI package, pinned in `pyproject.toml`/`uv.lock` — **not** a git/path dependency. Until this PR is merged to `main` **and** tagged (`pii-filter-v0.3.6`, which triggers the PyPI publish workflow), the gateway's `uv.lock` still resolves `cpex-pii-filter` to the last published `0.3.5`, which predates the metrics-emission behavior this whole effort adds. Concretely: `IBM/mcp-context-forge`'s new e2e test (`test_otel_plugin_metadata_e2e.py`) will not reliably pass in that repo's CI until the gateway's dependency is bumped to `>=0.3.6` after this package is published — see the merge-order note on PR #5470 for the exact sequence.


lucarlig:
Nice work — the metadata-vs-`Extensions.custom` rationale is well argued and the implementation matches it. I verified back-compat at both the shim and PyO3 signature levels, the trace-gated emission and S1 allow-list in code, and that `process_nested_internal` guarantees detections ⇒ modified, so no detection path can skip metrics. No new dependencies and no plugin-side telemetry transport — everything rides the result object / logging.

A few non-blocking suggestions:

- **Duplication:** the blocked-path result construction is ~30 near-identical lines in `prompt_post_fetch` (plugin.rs:126-153) and `handle_nested_stage` (288-315); a shared helper would collapse it.
- **Work before the gate:** `sorted_detection_types` + the `Vec<&str>` are built before `push_metrics_kwarg` checks `trace_id`, so untraced requests (the common case) pay the allocations — and since the caller gates, `build_pii_metrics`'s `trace_id.is_none()` / `Ok(None)` branches are unreachable. Passing the detections map into `push_metrics_kwarg` and collecting types after the gate fixes both.
- **Dead code:** `let _ = &context;` (plugin.rs:175) and the unused `_context` param on `handle_nested_stage` — the internal fn could drop the param entirely.
- **Contract wording:** on block, `prompt_post_fetch` returns at the first offending message, so `total_detections` is scan-until-block rather than whole-payload; the contract doc says "total count across all patterns in this hook invocation" — worth a clarifying sentence, since the gateway validator treats the doc as truth.
- **Minor:** `test_prompt_pre_fetch_records_metadata_and_logs_when_enabled` still passes `include_detection_details=True` (now a no-op), which slightly misleads about what gates metrics.
- **Follow-up, pre-existing:** the detector's opt-in debug log emits the nested field path, which the new contract's deny-list treats as sensitive for metrics — worth aligning in a later change.

Nothing blocking from my side.

lucarlig (approved):
LGTM

msureshkumar88:
Thanks for the review, @lucarlig — addressed the non-blocking suggestions in 76e351c:

- **Duplication**: extracted `build_blocked_result` to collapse the blocked-path construction shared by `prompt_post_fetch` and `handle_nested_stage`.
- **Work before the gate**: `push_metrics_kwarg` now checks `trace_id` before sorting/collecting detection types; `prompt_post_fetch` accumulates a `HashMap` across messages instead of pre-sorting per message, so untraced requests skip the allocation entirely.
- **Dead code**: dropped the unused `_context` param from `handle_nested_stage`; the three hook methods that only forwarded it to that helper keep `context` (required by the hook signature) but mark it unused via `let _ = &context;`.
- **Contract wording**: added a clarifying sentence to `pii_filter-metrics-schema.md` — `prompt_post_fetch`'s blocked path reports only the message that triggered the block, not the whole payload.
- **Minor**: removed the misleading `include_detection_details=True` from `test_prompt_pre_fetch_records_metadata_and_logs_when_enabled`.

Left the pre-existing detector debug-log/field-path item as-is, per your framing of it as a separate follow-up.

Verified before pushing: `cargo test --lib` (75 passing), `make test-integration` (26 passing), `cargo clippy --all-targets -- -D warnings` (clean), `cargo fmt --check` (clean).

lucarlig:
merging without s390x

PR #131, opened 2026-07-13
feat(plugins): emit result.metadata metrics from 5 remaining bundled plugins

### LINKED ISSUE #129: [FEATURE]: Emit result.metadata metrics from other plugins (secrets-detection, encoded-exfil-detection, url-reputation, rate-limiter, retry-with-backoff)
## Summary

The gateway's plugin-metrics pipeline (G1 internal DB spans + G2 optional OTel-SDK export, `IBM/mcp-context-forge#5458` / `#5470`) consumes `PluginResult.metadata[<plugin_name>]` from any plugin, generically. `pii_filter` (>=0.3.6) is the only plugin that currently populates it, so the pipeline has nothing to show for the rest of the bundled plugins.

Tracking issue on the gateway side: IBM/mcp-context-forge#5554

This issue tracks the plugin-side half of that work: emit `result.metadata[<plugin_name>] = {...}` from each plugin below, following the same contract `pii_filter` already uses.

### Contract (mirror `pii_filter`)

- Only scalar (`int`/`float`/`bool`) or short, low-cardinality `str`/`list[str]` fields -- counts and type/category names, never raw matched content, URLs, secrets, or free text (the gateway's S4 sanitizer will drop anything that doesn't fit a low-cardinality charset anyway, but don't rely on that as the privacy boundary -- the plugin must not put sensitive data in `metadata` in the first place)
- Gate emission on an active trace: read `extensions.request.trace_id` (as `pii_filter` does) and no-op if it's not set, so there's no overhead when observability is off
- Bump the plugin's own package version on release so the gateway repo can pin a floor version that includes metrics emission (see `IBM/mcp-context-forge#5470` for how `cpex-pii-filter`'s version floor was tightened for exactly this reason)

### Plugins in scope

- [ ] `cpex-secrets-detection` (`SecretsDetectionPlugin` -- `prompt_pre_fetch`, `tool_post_invoke`, `resource_post_fetch`): e.g. `total_detections`, `total_masked`/`total_blocked`, `secret_types` (category names only, e.g. `["aws_key", "api_token"]` -- never the matched value)
- [ ] `cpex-encoded-exfil-detection` (`EncodedExfilDetectorPlugin` -- `prompt_pre_fetch`, `tool_post_invoke`, `resource_post_fetch`): e.g. `total_detections`, `encoding_types` (e.g. `["base64", "hex"]`)
- [ ] `cpex-url-reputation` (`URLReputationPlugin` -- `resource_pre_fetch`): e.g. `total_checked`, `total_blocked`, `reputation_categories` (no raw URLs/domains)
- [ ] `cpex-rate-limiter` (`RateLimiterPlugin` -- `prompt_pre_fetch`, `tool_pre_invoke`): e.g. `allowed`, `throttled`, `backend` (redis/memory)
- [ ] `cpex-retry-with-backoff` (`RetryWithBackoffPlugin` -- `tool_post_invoke`): e.g. `retry_count`, `total_backoff_ms`

### Out of scope

- `pii_filter` (already done, 0.3.6)
- Plugins with no meaningful counters to surface

Refs IBM/mcp-context-forge#5458, IBM/mcp-context-forge#5470, IBM/mcp-context-forge#5554.

msureshkumar88:
## Summary

Closes #129. Extends the `result.metadata["<plugin_name>"]` OpenTelemetry metrics contract — already shipped for `pii_filter` (#128, merged) — to the 5 remaining bundled plugins: `secrets_detection`, `rate_limiter`, `retry_with_backoff` (Rust-core), and `encoded_exfil_detection`, `url_reputation` (Python-native).

Each plugin now accepts an optional `extensions` param on its in-scope gateway hooks, reads `extensions.request.trace_id`, and — only when a trace is active — emits a namespaced, allow-listed metrics dict (counts/type-category names only, never matched content) to `result.metadata["<plugin_name>"]`. Untraced calls are behavior-identical to before this change.

**Not a greenfield rollout.** 4 of the 5 plugins already wrote some form of `result.metadata` before this PR — flat/non-namespaced, un-gated, and in `encoded_exfil_detection`'s case, with a real data-leak (S1) risk via raw finding content. This PR migrates each to the new namespaced/gated contract, removing the old writes rather than adding alongside them (no double-write). Only `url_reputation` was greenfield.

## What changed per plugin

| Plugin | Hooks in scope | Version | Notes |
|---|---|---|---|
| `secrets_detection` | `prompt_pre_fetch`, `tool_post_invoke`, `resource_post_fetch` | 0.3.6 → 0.3.7 | Migrated: removed flat `secrets_redacted`/`secrets_findings` keys |
| `rate_limiter` | `prompt_pre_fetch`, `tool_pre_invoke` | 0.1.6 → 0.1.7 | Migrated: previously-metadata-less throttled branch now emits metrics too |
| `retry_with_backoff` | `tool_post_invoke` only | 0.3.5 → 0.3.6 | Migrated: dropped un-gated `retry_policy` echo. Emits `retry_delay_ms` (per-attempt) — no cumulative `total_backoff_ms`, no new state accumulator added |
| `encoded_exfil_detection` | `prompt_pre_fetch`, `tool_post_invoke`, `resource_post_fetch` | 0.3.5 → 0.3.6 | Migrated: closed a real S1 leak (raw finding content) in the old flat write |
| `url_reputation` | `resource_pre_fetch` only | 0.3.4 → 0.3.5 | Greenfield. `reputation_categories` uses this plugin's actual violation-reason slugs |

Also adds `docs/superpowers/contracts/<plugin>-metrics-schema.md` for all 5, documenting actual shipped fields (not always the same as this issue's original illustrative examples).

## Known follow-up (not a blocker)

`encoded_exfil_detection` emits no metrics at all on its blocking path (unlike `secrets_detection`/`url_reputation`, which do) — disclosed in its contract doc. Tracked as a follow-up to align block-path telemetry across plugins, not fixed in this PR.

## Test plan

- [x] Trace-in/metrics-out coverage for every plugin/hook in scope
- [x] No-extensions back-compat (legacy callers still work)
- [x] S1 leak-check tests (no raw secret/finding/URL content ever reaches `result.metadata`)
- [x] Legacy-key regression tests (old flat keys confirmed gone)
- [x] `make ci` green in all 5 plugin directories


plugins/rust/python-package/encoded_exfil_detection/cpex_encoded_exfil_detection/encoded_exfil_detection.py
```diff
@@ -211,16 +239,23 @@ async def prompt_pre_fetch(self, payload: PromptPrehookPayload, context: PluginC
                 ),
             )
 
-        metadata = {"encoded_exfil_count": count, "encoded_exfil_findings": self._findings_for_metadata(findings), "implementation": self.implementation} if count else {}
+        metrics = self._build_metrics(extensions, count, findings)
```
lucarlig: **[P1] Emit metrics on the blocking path.** Metric construction occurs after the blocking return. With the default `block_on_detection=True` and threshold `1`, every actual detection exits before metadata is attached, while clean calls emit nothing; this plugin therefore produces no metrics under its default configuration. Please build metrics before the condition and attach them to blocked results in all three hooks, with traced-block regression tests.

plugins/rust/python-package/rate_limiter/src/plugin.rs
```diff
@@ -366,42 +394,127 @@ fn build_prehook_result(
...
+    build_framework_object_dyn(py, class_name, kwargs)
+}
+
+/// `meta` fields (built by `engine::build_meta_dict`) that are safe to fold
+/// into the namespaced metrics dict: aggregate rate-limit state only.
+/// Deliberately excludes `user_id`/`tenant_id` — `build_meta_dict` only sets
+/// those on the not-allowed path (G7, for `PluginViolation.details`
+/// debugging), and a metrics/telemetry channel is not the place for identity
+/// (S1: no identifiers in metrics, mirrors "no raw content").
+const META_METRIC_KEYS: &[&str] = &["limited", "remaining", "reset_in", "dimensions"];
```
lucarlig: **[P2] Keep rate-limiter metadata within the consumer contract.** `dimensions` is a nested dict, but issue #129 and the gateway sanitizer permit only scalars or `list[str]`, so it is always dropped. The companion gateway PR #5611 also does not allowlist numeric `remaining` or `reset_in`; those fields are silently discarded too. Please emit only the supported `allowed`/`throttled`/`backend` contract (and any deliberately supported bool), or flatten and coordinate the additional fields with the consumer and its tests.

plugins/rust/python-package/rate_limiter/src/bin/stub_gen.rs
```diff
@@ -27,7 +27,7 @@ fn curate_extension_stub() {
     );
     if !content.contains("class RateLimiterPluginCore:") {
         content.push_str(
-            "\n\n@typing.final\nclass RateLimiterPluginCore:\n    def __new__(cls, config: dict) -> RateLimiterPluginCore: ...\n    def prompt_pre_fetch(self, payload: typing.Any, context: typing.Any) -> typing.Any: ...\n    def tool_pre_invoke(self, payload: typing.Any, context: typing.Any) -> typing.Any: ...\n",
+            "\n\n@typing.final\nclass RateLimiterPluginCore:\n    def __new__(cls, config: dict) -> RateLimiterPluginCore: ...\n    def prompt_pre_fetch(self, payload: typing.Any, context: typing.Any, extensions: typing.Any = None) -> typing.Any: ...\n    def tool_pre_invoke(self, payload: typing.Any, context: typing.Any, extensions: typing.Any = None) -> typing.Any: ...\n",
```
lucarlig, line 30: **[P2] Regenerate the shipped PyO3 stubs.** The generator template knows about `extensions`, but the checked-in rate-limiter stub still advertises two-argument hook methods; the retry-with-backoff and secrets-detection stubs are stale in the same way. Type checkers will reject valid calls against the released wheels. `make stub-gen` currently fails on this checkout because `pyo3-stub-gen 0.22.3` is incompatible with `pyo3 0.29.0`, so this will not repair itself during release. Please fix the generator dependency and regenerate all three stubs, or update them explicitly and add a generated-diff verification step.

plugins/rust/python-package/secrets_detection/src/plugin.rs
```diff
@@ -54,30 +60,40 @@ impl SecretsDetectionPluginCore {
             "args",
             "ToolPreInvokeResult",
             "Potential secrets detected in tool arguments",
+            None,
```
lucarlig: **[P2] Preserve telemetry for `tool_pre_invoke`.** This hook remains advertised in the plugin manifest and previously emitted findings/redaction metadata, but passing `None` here guarantees it now emits neither legacy nor namespaced metrics, even when the caller has an active trace. That leaves existing users with no replacement path. Please accept optional `extensions` and emit the same namespaced schema here, or preserve the legacy metadata behavior for this explicitly out-of-scope hook.

lucarlig (changes requested):
Suresh, I found one default-path blocker and three contract/packaging regressions that should be resolved before merge.

The main blocker is that `encoded_exfil_detection` emits no metrics under its default blocking configuration: every actual finding returns before metric construction, while clean calls intentionally emit nothing. I also found a rate-limiter schema mismatch with the gateway consumer, stale shipped PyO3 stubs for the new method signatures, and a permanent metadata regression on `secrets_detection.tool_pre_invoke`.

Verification performed against this head/code revision:
- 217 Rust tests passed.
- 376 Python integration tests passed; 2 expected xfails.
- `cargo fmt --all --check`, package-target Clippy, and `git diff --check` passed.
- `make stub-gen` failed because `pyo3-stub-gen 0.22.3` is incompatible with `pyo3 0.29.0`, which makes the stale checked-in stubs especially important to address.

Please see the inline comments for concrete fixes and regression-test suggestions.

plugins/rust/python-package/encoded_exfil_detection/cpex_encoded_exfil_detection/encoded_exfil_detection.py
msureshkumar88: Confirmed — default `block_on_detection=True` + `min_findings_to_block=1` meant every real detection returned before `_build_metrics()` ran, so metrics never fired under default config. Fixed in 45a6e3b: metrics are now built before the blocking check in all 3 hooks and attached to the blocked `PluginViolation` result too (still only `total_detections`/`encoding_types`, no raw findings — S1 unaffected). Added 4 regression tests covering the default-config blocking path for all 3 hooks.

plugins/rust/python-package/rate_limiter/src/plugin.rs
msureshkumar88: Agreed — `dimensions` is a nested dict and can never pass the gateway's scalar/list[str]-only sanitizer, and `remaining`/`reset_in` weren't on its numeric allowlist either. Fixed in b52b319: dropped `limited`/`remaining`/`reset_in`/`dimensions` entirely, metrics dict now only carries `allowed`/`throttled`/`backend` (the fields actually in the gateway's contract). Updated Rust + Python tests to assert the key set is exactly that triple, plus README/contract doc.

plugins/rust/python-package/rate_limiter/src/bin/stub_gen.rs
msureshkumar88: Confirmed and reproduced the `pyo3-stub-gen` 0.22.3 / pyo3 0.29.0 incompatibility exactly as you described. Found 0.23.0 added pyo3 0.29 support, bumped it, and `make stub-gen` then regenerated cleanly for all 3 plugins (rate_limiter, retry_with_backoff, secrets_detection) — no manual stub hand-editing needed. Fixed in fed7d21 (rate_limiter, retry_with_backoff) and 11bff5a (secrets_detection, alongside its own signature change from the tool_pre_invoke fix). Regen also caught that the other hooks' stubs had drifted on a stale `_context` param name.

plugins/rust/python-package/secrets_detection/src/plugin.rs
msureshkumar88: Confirmed — this was a real regression, not just an out-of-scope omission: the hook lost its old legacy metadata during the RECONCILE cleanup and got nothing in return. Fixed in 11bff5a: `tool_pre_invoke` now takes `extensions`, wires `read_trace_id` into `scan_payload_attr` the same way the other 3 hooks do, and emits the same namespaced/gated schema. Also found and fixed a bug this would have masked: the Python `SecretsDetectionPlugin` shim was still calling this hook with only 2 args, so the Rust-side fix alone would never have been reachable end-to-end. Updated the "out of scope" language in README/contract doc, added a regression test.

msureshkumar88:
Thanks for the thorough review — all 4 findings confirmed valid and fixed (replied inline on each thread with the specific commit):

- **P1 (encoded_exfil_detection default-path blocker)**: `45a6e3b` — metrics now built before the blocking check, attached to blocked results too.
- **P2 (rate_limiter schema mismatch)**: `b52b319` — dropped the fields the gateway sanitizer can't structurally accept (`dimensions`, `remaining`, `reset_in`); metrics now only carry `allowed`/`throttled`/`backend`.
- **P2 (stale PyO3 stubs)**: `fed7d21` + `11bff5a` — bumped `pyo3-stub-gen` 0.22.3→0.23.0 (adds pyo3 0.29 support) and regenerated all 3 affected stubs properly via `make stub-gen` rather than hand-editing.
- **P2 (secrets_detection tool_pre_invoke regression)**: `11bff5a` — hook now accepts `extensions` and emits the same schema as the other 3 hooks. Also caught and fixed a related bug: the Python shim was still calling this hook with only 2 args, which would've made the Rust fix unreachable.

Full verification (Rust `cargo test`/clippy/fmt, Python integration suites) before each push, plus new regression tests per finding. CI is green on the latest commit. Ready for another look whenever convenient.

lucarlig (approved):
Thanks, Suresh — I re-reviewed the fixes on 11bff5a. All four requested changes are implemented: blocking-path metrics are emitted for encoded-exfil detection, rate-limiter metadata now matches the gateway contract, the PyO3 generator dependency and shipped stubs are current, and secrets_detection.tool_pre_invoke preserves traced telemetry end-to-end. Focused verification passed (20 Python regression/integration tests, 163 Rust tests, rustfmt, Clippy, and clean stub regeneration), and CI is green. Approved.

Issue #136, opened 2026-07-14
RetryWithBackoffPlugin (Rust, >=0.3.2) never retries in the gateway: is_failure reads CopyOnWriteDict via C-level PyDict::get_item


msureshkumar88:
## Summary

`RetryWithBackoffPlugin` (Rust core, `>=0.3.2`) never signals a retry when run inside the MCP ContextForge gateway. Every failing tool result is treated as a success, so the gateway re-invokes nothing. This is a regression introduced by the Python→Rust rewrite in `0.3.2`; the pure-Python `0.3.1` works correctly.

## Impact

- The plugin's core function (retry on transient tool failures) is silently broken in the gateway.
- The OTel retry metrics (`retry_count`, `retry_delay_ms`) are always zero, since `is_failure` never fires.
- Gateway E2E `tests/live_gateway/plugins/test_retry_with_backoff_e2e.py` fails deterministically (both `static` and `binding` legs) whenever the gateway pins `>=0.3.2`. This is why `IBM/mcp-context-forge#5332` pinned `cpex-retry-with-backoff==0.3.1`, and why `IBM/mcp-context-forge#5611` (which bumps to `>=0.3.6`) is red.

## Root cause

The gateway framework wraps every `tool_post_invoke` result payload in `cpex.framework.memory.CopyOnWriteDict` for copy-on-write payload isolation. `CopyOnWriteDict` is a `dict` **subclass** that keeps its data in a Python-level overlay and leaves the C-level `dict` storage **empty**:

```python
d = CopyOnWriteDict({"isError": True})
dict.__len__(d)         # 0   (C-level storage empty)
len(d)                  # 1   (overridden)
dict.get(d, "isError")  # missing at C level
d["isError"]            # True (overridden __getitem__ reads the overlay)
```

`is_failure` in `retry_with_backoff/src/plugin.rs` reads keys with `PyDict::get_item`, which accesses the **C-level** dict storage directly (`PyDict_GetItem`) and therefore observes an empty dict — it never finds `isError`, returns `false`, and the plugin reports `retry_delay_ms=0` (its "not a failure" path). The gateway uses `retry_delay_ms > 0` as its retry signal, so it never retries.

`0.3.1` (Python) used `result.get("isError")`, which goes through the overridden `__getitem__`/`get` and reads the overlay correctly.

### Reproduced end-to-end
- Real gateway + real fast-time-server (pinned image), instrumented log at the retry check:
[...]

PR #137, opened 2026-07-14
fix(retry_with_backoff): read result via mapping protocol so CopyOnWriteDict works (#136)

### LINKED ISSUE #124: retry-with-backoff (Rust-majority) fails to detect failures when gateway isolates the payload (CopyOnWriteDict) — no retries
## Summary

The Rust-majority `RetryWithBackoffPlugin` (>= 0.3.2, after the migration in #91 `feat(retry): migrate to Rust-majority`) does **not** detect tool failures when the host gateway isolates the `tool_post_invoke` payload. As a result, `is_failure` returns `false`, `retry_delay_ms` is `0`, and **no retry is ever attempted** — the gateway makes only a single upstream call.

This is a behavioral regression vs. the pure-Python 0.3.1 plugin, which worked correctly.

## Impact

- Plugin-integration E2E tests fail in `mcp-context-forge` (the `retry-with-backoff` static + binding legs):
  - `test_transient_failures_are_retried_until_success`
  - `test_failures_beyond_budget_surface_error`
- In production, a transient upstream failure (`isError: true`) is silently passed through instead of being retried.

## Root cause

When a host like `mcp-context-forge` registers a `tool_post_invoke` hook that declares writable fields, the cpex framework isolates the payload via `cpex.framework.memory.wrap_payload_for_isolation`. This converts `payload.result` from a plain `dict` into a **`CopyOnWriteDict`** (a `dict` subclass).

`CopyOnWriteDict.__init__` calls `super().__init__()` with **no data**, leaving the underlying C-level dict storage empty. The real data lives in `self._original` and is exposed only through **Python-level overrides** (`__getitem__`, `get`, `__contains__`, `__iter__`, `__len__`).

The Rust `is_failure` (in `plugins/rust/python-package/retry_with_backoff/src/plugin.rs`) reads keys like this:

```rust
let Ok(result_dict) = result.cast::<PyDict>() else {
    return Ok(false);
};
if let Some(is_error) = result_dict.get_item("isError")? // <-- C-level read
    && is_error.extract::<bool>().unwrap_or(false)
{ ... }
```

pyo3's `PyDict::get_item` reads the **C-level** dict storage directly (`PyDict_GetItem`), bypassing the Python-level override. Against a `CopyOnWriteDict`, the C-level storage is empty, so:

- `get_item("isError")` → `None`
- `is_failure` → `Ok(false)` ("not a failure")
- `retry_delay_ms` → `0` → **no retry**

The pure-Python 0.3.1 implementation used `result.get("isError")`, which dispatches to the override and therefore worked.

## Reproduction (verified locally)

- Plain `dict {isError: true}` through the plugin → `retry_delay_ms = 200` (correct).
- `wrap_payload_for_isolation(payload)` → `result` is a `CopyOnWriteDict` → plugin → `retry_delay_ms = 0` (bug).
- For `d = CopyOnWriteDict({'isError': True})`:
  - Python-level: `d.get('isError') == True`, `'isError' in d == True`, `len(d) == 2`
  - C-level: `dict.get(d, 'isError')` → missing, `list(dict.__iter__(d)) == []`
- Live gateway instrumentation confirmed: `has_post=True`, `is_error=True`, `result` dump contains `isError: True`, `max_retries=2` — yet the plugin returns `retry_delay_ms=0` and the Rust `debug!("tool_post_invoke: tool=...")` line never appears (it takes the silent `!is_failure` branch).

## Suggested fix

Read dict keys in a way that honors `dict` *subclass* overrides — mirroring how `secrets_detection` already handles this. In `secrets_detection/src/scanner/python_scan.rs` it branches on exact-vs-subclass:

```rust
if container.is_exact_instance_of::<PyDict>() {
    // fast C-level path
    let dict = container.cast::<PyDict>()?;
    for (_, value) in dict.iter() { ... }
} else if container.is_instance_of::<PyDict>() {
    // dict subclass: dispatch to Python so overrides apply
    for item in container.call_method0("items")?.try_iter()? { ... }
}
```

For `retry_with_backoff`, apply the same idea to every key read in `is_failure` (`isError`, `structuredContent`, `status_code`, `content`, and the nested content-item reads): use the C-level `get_item` only for an *exact* `dict`, and fall back to a Python-level `result.call_method1("get", (key,))` for subclasses so the override is honored.

## Affected code

- `plugins/rust/python-package/retry_with_backoff/src/plugin.rs` — `is_failure`

## Notes

- Other Rust plugins that read mapping payloads (e.g. anything using `get_item` on `result`/`args`) should be audited for the same C-level-read pitfall.
- Suggested test: add a unit test that runs `is_failure` against a `dict` subclass whose data is only reachable via Python-level `get`/`__getitem__` (i.e. an empty C-level store), asserting the failure is still detected.

msureshkumar88:
## Related issue

Closes #124

## Problem

`RetryWithBackoffPlugin` (Rust core, `>=0.3.2`) never retries inside the MCP ContextForge gateway — every failing tool result is treated as a success. Root cause: the gateway wraps `tool_post_invoke` results in `cpex.framework.memory.CopyOnWriteDict`, a `dict` subclass that keeps its data in a Python-level overlay and leaves the C-level dict storage empty. `is_failure` read keys with `PyDict::get_item` (C-level `PyDict_GetItem`), so it observed an empty dict, never found `isError`, and returned `retry_delay_ms=0`. The gateway retries only when `retry_delay_ms > 0`, so it never retried.

Regressed in the `0.3.2` Python→Rust rewrite; pure-Python `0.3.1` used `result.get(...)` (the overridden `__getitem__`) and worked. Full analysis and reproduction in #136.

## Fix

- Read every result key through the Python mapping protocol (`PyObject_GetItem` → `__getitem__`) via new `mapping_get` / `mapping_get_bool` helpers, instead of `PyDict::get_item`. Dict subclasses with overlay storage now work; plain-dict behaviour is unchanged.
- Regression test `test_is_failure_reads_dict_subclass_overlay_storage`: feeds `is_failure` a dict subclass whose C-level storage is empty and asserts the failure is still detected (asserts C-level `len == 0` first, so it would fail against the old code).
- Bump `0.3.6` → `0.3.7`; fix the stale `v0.3.2` init log string.

## Verification

- `cargo test --lib` — 56/56 pass (incl. the new regression test).
- `cargo fmt --check`, `cargo clippy --all-targets` — clean.
- Rebuilt the wheel and ran the gateway E2E (`tests/live_gateway/plugins/test_retry_with_backoff_e2e.py`) end-to-end against the real fast-time-server image: **both `static` and `binding` legs pass** (they fail against the current `0.3.6`).

## Downstream

`IBM/mcp-context-forge#5611` will pin `cpex-retry-with-backoff>=0.3.7` once this ships to PyPI.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1565,7 +1565,7 @@ checksum = "d6f6ff9a378485b298a5286656da665ba74413d36db0979633275d2e708145d4"
 
 [[package]]
 name = "retry_with_backoff"
-version = "0.3.6"
+version = "0.3.7"
 dependencies = [
  "cpex_framework_bridge",
  "log",
--- a/plugins/rust/python-package/retry_with_backoff/Cargo.toml
+++ b/plugins/rust/python-package/retry_with_backoff/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "retry_with_backoff"
-version = "0.3.6"
+version = "0.3.7"
 edition.workspace = true
 authors.workspace = true
 license.workspace = true
--- a/plugins/rust/python-package/retry_with_backoff/cpex_retry_with_backoff/plugin-manifest.yaml
+++ b/plugins/rust/python-package/retry_with_backoff/cpex_retry_with_backoff/plugin-manifest.yaml
@@ -1,6 +1,6 @@
 description: "High-performance retry policy engine with exponential backoff, jitter, per-tool overrides, and retry metadata for transient tool and resource failures - Rust-majority architecture (95% Rust / 5% Python)"
 author: "ContextForge Contributors"
-version: "0.3.6"
+version: "0.3.7"
 kind: "cpex_retry_with_backoff.retry_with_backoff.RetryWithBackoffPlugin"
 available_hooks:
   - "tool_post_invoke"
--- a/plugins/rust/python-package/retry_with_backoff/cpex_retry_with_backoff/retry_with_backoff.py
+++ b/plugins/rust/python-package/retry_with_backoff/cpex_retry_with_backoff/retry_with_backoff.py
@@ -50,7 +50,7 @@ class RetryWithBackoffPlugin(Plugin):
                     override["max_retries"] = ceiling
 
         self._core = RetryWithBackoffPluginCore(raw_cfg)
-        log.info("retry_with_backoff: Initialized with Rust core (v0.3.2)")
+        log.info("retry_with_backoff: Initialized with Rust core (v0.3.7)")
 
     async def tool_post_invoke(
         self,
--- a/plugins/rust/python-package/retry_with_backoff/cpex_retry_with_backoff/retry_with_backoff.py
+++ b/plugins/rust/python-package/retry_with_backoff/cpex_retry_with_backoff/retry_with_backoff.py
@@ -16,6 +16,7 @@ from cpex.framework import (
 )
 from cpex.framework.settings import get_settings
 
+from cpex_retry_with_backoff.retry_with_backoff_rust import __version__ as _RUST_CORE_VERSION
 from cpex_retry_with_backoff.retry_with_backoff_rust import RetryWithBackoffPluginCore
 
 log = logging.getLogger(__name__)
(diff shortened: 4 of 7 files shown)


lucarlig (changes requested):
@msureshkumar88 Please make `mapping_get` distinguish missing keys from unexpected `__getitem__` errors; `.ok()` currently turns every mapping failure into “not a failure,” risking another silent no-retry path. Also remove or derive the hard-coded `v0.3.7` log version so it does not go stale again. The mapping-protocol fix otherwise looks good.

lucarlig (approved):
LGTM

PR #142, opened 2026-07-21
fix(secrets_detection): preserve secret field names during redaction

lucarlig:
## Summary

- redact only the detected value for `aws_secret_access_key` assignments, preserving the field name, separator, whitespace, and quotes
- apply the same value-only replacement contract to `generic_api_key_assignment`
- strengthen exact-output regression coverage for AWS assignment formats and every supported detector
- bump `secrets_detection` to 0.3.10

## Root cause

The AWS detector matched the full labeled assignment and captured the value, but the scanner only used capture-group spans for the base64 detector. Redaction therefore replaced the detector's full match, including the field name. The generic API-key assignment detector had the same whole-assignment replacement behavior and did not expose a value capture.

## Impact

An input such as:

```text
Aws Secret Access Key = "wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY"
```

is now returned as:

```text
Aws Secret Access Key = "[REDACTED]"
```

instead of replacing the entire assignment.

## Validation

- `make ci` in `plugins/rust/python-package/secrets_detection`
- 96 Rust tests passed
- 114 plugin-framework integration tests passed
- formatting, Clippy with warnings denied, wheel build/install, and benchmark compilation passed

## Tracker

Addresses the follow-up reported in [WatsonOrchestrate/wo-tracker#78411](https://github.ibm.com/WatsonOrchestrate/wo-tracker/issues/78411#issuecomment-223697436).

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "(^\\.secrets\\.baseline$|package-lock\\.json$|Cargo\\.lock$|uv\\.lock$|go\\.sum$|poetry\\.lock$|Pipfile\\.lock$)|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2026-07-15T10:10:56Z",
+  "generated_at": "2026-07-21T09:27:16Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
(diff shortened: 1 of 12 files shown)


gandhipratik203 (approved):
Clean fix, and the exact-output tests are a nice upgrade. One non-blocking nit: the README Notes only mention capture-group redaction for `base64_24`, but this now applies to `aws_secret_access_key` and `generic_api_key_assignment` too — worth a one-line doc update.

PR #143, opened 2026-07-21
fix: release newly added plugins

lucarlig:
## Summary

- Treat a newly added plugin `Cargo.toml` as an initial release candidate.
- Add regression coverage for first-time plugin additions.

## Root cause

Release tag detection only emitted a tag when the plugin had an existing base version that changed. For a new plugin, the base `Cargo.toml` does not exist, so `release_validation_tags` was empty and both release validation and tag creation were skipped.

## Impact

New plugins will now follow the same validation, automatic tagging, and release path as plugin version bumps.

## Validation

- `make plugins-validate` — 125 tests passed, 3 skipped.
- Focused initial-release and version-bump tests pass.
- Replayed the SQL sanitizer merge diff; it now emits `sql-sanitizer-v0.1.0` with `has_release_validation_tags: true`.

### DIFF
--- a/tests/test_plugin_catalog.py
+++ b/tests/test_plugin_catalog.py
@@ -1737,6 +1737,43 @@ class PluginCatalogTests(unittest.TestCase):
             self.assertEqual(payload["release_validation_tags"], ["rate-limiter-v0.0.2"])
             self.assertTrue(payload["has_release_validation_tags"])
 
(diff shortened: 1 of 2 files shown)


madhu-mohan-jaishankar (approved):
LGTM

gandhipratik203 (approved):
LGTM

PR #144, opened 2026-07-21
fix(ci): preserve release-bearing main runs

lucarlig:
## What changed

- cancel superseded `CI Rust Python Package Plugins` runs only for pull requests
- keep `main` runs serialized without cancelling the run already in progress
- dispatch releases from the immutable release tag instead of the moving `main` ref
- add a regression test for the concurrency policy

## Why

Commit `760a223c35fa6150bb2789b345ff422cf9b1028b` bumped `secrets_detection` to `0.3.10`, but its `main` CI run was cancelled when the following Dependabot merge started another run in the same concurrency group. The cancelled run never reached `create-release-tags`, so the release could not be dispatched.

Version bumps are detected only after the workflow starts. Allowing every `main` run to finish ensures a release-bearing commit completes the full CI and reaches tag creation, while superseded pull-request runs remain cancellable.

The release dispatcher also used `--ref main`. If another commit advanced `main` while the release-bearing CI run finished, the release workflow's shallow checkout could not prove the older tag was on `main` and skipped the publish job. Dispatching from `${tag}` keeps the workflow ref pinned to the validated release commit and makes the ancestry guard reliable.

## Impact

- release-bearing `main` runs continue through full CI and release dispatch
- subsequent `main` runs wait for the active run instead of cancelling it
- release builds and publishing stay pinned to the created tag even if `main` advances
- pull-request concurrency behavior is unchanged

## Validation

- `python3 -m unittest tests/test_plugin_catalog.py` (112 passed, 3 skipped)
- pre-commit YAML, whitespace, merge-conflict, and secret checks
- shallow-checkout ancestry probe using `secrets-detection-v0.3.10`


msureshkumar88 (approved):
## E2E verification of the fix

I couldn't trigger real GitHub Actions runs (no way to script GH's scheduler from outside, and doing so would mutate CI/release state on the repo), so I verified the two behavioral claims the way they *can* be verified for real: by executing the actual shipped logic locally, plus a faithful state-machine simulation of GitHub's documented concurrency-queueing rules.

### Config
- Cloned `user/luca/preserve-release-ci` (HEAD `38b939b`, contains latest `main`), shallow, into an isolated sandbox — no changes pushed anywhere.
- Python 3.12.3, bash 5, real local git remote (`git init --bare`) standing in for GitHub, plus a mocked `gh` binary on `PATH` that logs call order instead of hitting the network.

### 1. Real test suite (matches PR's own "Validation" section)
```
python3 -m unittest tests/test_plugin_catalog.py -v
```
Result: **112 passed, 2 skipped** (PR description says 3 skipped locally — env-dependent skip, not a regression).

### 2. Concurrency-queueing simulation
Modeled GitHub's documented group/cancel-in-progress queueing rules and drove them with the actual `group:`/`cancel-in-progress:` expressions parsed straight out of the PR's YAML, replaying the real incident timeline from the PR description (commit `760a223` release-bearing push, immediately followed by the dependabot `624b2e2` push into the same `main` concurrency group):

- **Old config** (`cancel-in-progress: true`): run-A (release-bearing) → CANCELLED, run-B takes over. Reproduces the reported bug.
- **New config** (`cancel-in-progress: ${{ github.event_name == 'pull_request' }}`): run-A → stays IN_PROGRESS to completion, run-B queues PENDING and starts only after. Bug fixed.
- Regression check: two pushes to the same PR branch still cancel-supersede correctly (`pull_request` event ⇒ `cancel-in-progress: true`) — no change to PR behavior.

### 3. Sandbox execution of the actual `create-release-tags` step
Extracted the literal `run:` block for "Create release tags for bumped plugin versions" byte-for-byte from the workflow file and ran it under `bash -euo pipefail` against a real git remote + mocked `gh`:

- Tag `secrets-detection-v0.3.10` created and `git push origin refs/tags/...` **completed before** `gh workflow run release-rust-python-package.yaml --ref "${tag}"` fired (confirmed via call-order log) — dispatch is never racing an un-pushed tag, and it targets the tag, never `main`.
- Re-running with the same tag/sha: skips tag creation, still dispatches (idempotent retry path intact).
- Same tag name at a different sha: aborts with `FATAL: tag ... already exists at ..., expected ...` (exit 1) — doesn't silently overwrite.
- Injection attempt (`v1.0; rm -rf /tmp/pwned`) as a tag value: rejected by the `^[a-zA-Z0-9._-]+$` guard before reaching any `git`/`gh` call.

### 4. Surrounding blast-radius check
- Other workflows (`ci-scaffold-generator.yaml`, `ci-plugin-catalog.yaml`, `ci-install-built-wheel.yaml`) still use `cancel-in-progress: true` — confirmed none of them run `create-release-tags` or dispatch releases, so they don't share this bug and are correctly out of scope.
- `release-rust-python-package.yaml`'s ancestry check (`git merge-base --is-ancestor "${tag_ref}" refs/remotes/origin/main`) does its own `git fetch --force` for both `main` and the tag ref before checking, so dispatching by tag (vs `main`) doesn't starve it of history — the shallow-checkout ancestry guard this PR's description mentions checks out.

### Verdict
All checks pass, no regressions found in the surrounding workflows. **Approving.**

PR #159, opened 2026-07-30
fix(sql_sanitizer): analyse executable comments, join root uv workspace

madhu-mohan-jaishankar:
## 🔗 Related Issue

Closes #

---

## 📝 Summary

Two changes to `sql_sanitizer`, plus the repo-structure cleanup they depend on.

**1. SQL hidden in executable comments was not analysed.**

MySQL and MariaDB *execute* the body of `/*! … */` (and MariaDB also `/*M! … */`),
while every other engine treats it as an ordinary comment. `strip_sql_comments()`
discarded it wholesale before detection ran, so a payload like:
`SELECT 1 /*!32302 ; DROP TABLE users */`

was analysed as a bare `SELECT 1` and passed the guard, while MySQL would run the
`DROP`. The same held for `DELETE`/`UPDATE` without `WHERE`.

New `unwrap_exec_comments()` rewrites executable comments to their body *for
analysis only*, so detection sees the statements the engine would actually run.
It is deliberately **not** used when rebuilding the outgoing payload — there the
comment is still stripped entirely, so the hidden SQL reaches no engine. Ordinary
block/line comments and quoted literals are passed through untouched (an
apostrophe inside `-- it's fine` would otherwise be read as opening a string
literal and hide a following executable comment).

**2. Comment stripping only rewrote top-level fields.**

`scan_args` recorded stripped replacements keyed by leaf name and applied them as a
shallow top-level overlay, so values nested in dicts or lists were *analysed* but
never *rewritten* — the original text was forwarded downstream. `StrippedFields`
now carries the full path (`Key`/`Index`) to each rewritten value, and
`rebuild_args_with_stripped` walks that path, copying containers as it descends
rather than mutating the caller's payload. The SQL sent downstream now matches the
SQL that was analysed, at any depth.

**3. `sql_sanitizer` joined the root uv workspace.**

It was the only plugin absent from `[tool.uv.workspace] members`, so `uv` treated it
as a standalone project with its own `uv.lock`. That lock resolved outside the root
`[tool.uv]` block, inheriting neither `exclude-newer = "10 days"` nor
`constraint-dependencies` — 14 of 67 shared packages had drifted from the root
resolution, and the root `maturin<2.0` ceiling did not reach its dev group (which
declares an uncapped `maturin>=1.13.3`, so maturin 2.0 would have broken this plugin
alone). Added to the workspace, standalone lock removed, and the special-casing it
required deleted: the second Dependabot entry and the extra `uv lock --upgrade` step
in `plugin-maintenance.yaml`.

Plugin version bumped `0.1.0` → `0.1.1` per the repo versioning convention.

---

## 📏 Reviewability

- [ ] This PR has one clear purpose
- [x] The linked issue is not labeled `triage`
- [x] Unrelated bugs or improvements are tracked in separate issues/PRs
- [x] Tests are included with the code they validate
- [x] If AI-assisted, I understand and can explain the generated changes

> Note on the first box: this PR intentionally bundles the `sql_sanitizer` fix with
> the uv-workspace change. They were combined so the plugin's dependency resolution
> and the fix ship together; reviewers who prefer them split should say so and the
> workspace change can be lifted into its own PR.

---

## 🏷️ Type of Change

- [x] Bug fix
- [ ] Feature / Enhancement
- [ ] Documentation
- [ ] Refactor
- [x] Chore (deps, CI, tooling)
- [ ] Other (describe below)

---

## 🧪 Verification

All commands run from `plugins/rust/python-package/sql_sanitizer` unless noted.

| Check | Command | Status |
|---|---|---|
| Full plugin CI | `make ci` | ✅ pass |
| Rust unit tests | `cargo test -p sql_sanitizer --lib` | ✅ 66 passed, 0 failed |
| Python integration tests | `make test-integration` | ✅ 17 passed |
| Formatting | `cargo fmt -- --check` | ✅ clean |
| Lint | `cargo clippy -p sql_sanitizer --all-targets -- -D warnings` | ✅ 0 warnings |
| Workspace-wide Rust tests | `cargo test --workspace --lib` (repo root) | ✅ all suites pass |
| Workspace re-lock | `uv lock` then `uv sync` (repo root) | ✅ no dependency versions changed |
| Plugin sync as member | `make sync` | ✅ resolves from root workspace |
| Coverage ≥ 90% | `tools/plugin_catalog.py coverage-check` | ⬜ not run locally — enforced by the `coverage` job in `ci-rust-python-package.yaml` |

15 tests added. One existing test changed meaning: `nested_comment_stripping_does_not_overwrite_top_level_key` asserted `modified_payload` is `None` for a nested comment — that assertion encoded the bug being fixed. It is now `nested_comment_stripping_rewrites_nested_field_only`, asserting the nested field *is* rewritten while an unrelated top-level key of the same name is left untouched. The original intent is preserved; only the buggy half is inverted.

Root lock diff contains no version changes — only the materialised `exclude-newer`
timestamp and the added `cpex-sql-sanitizer` member.

---

## ✅ Checklist

- [x] Code formatted (`cargo fmt`, `cargo clippy -- -D warnings`)
- [x] Tests added/updated for changes
- [ ] Documentation updated (if applicable) — no user-facing config or behaviour change; rationale documented in module doc comments
- [x] No secrets or credentials committed

---


lucarlig (approved):
LGTM

PR #169, opened 2026-08-24
Feat/output length guard rust migration

### LINKED ISSUE #145: Migrate output_length_guard plugin to Rust (from gateway Python)
## Summary

`output_length_guard` today lives only as pure Python in `mcp-context-forge` (`plugins/output_length_guard/`). Every other guard-style plugin (`pii_filter`, `secrets_detection`, `rate_limiter`, `retry_with_backoff`, `url_reputation`, `encoded_exfil_detection`) has already been migrated into `cpex-plugins` as a Rust core with thin PyO3 bindings. `output_length_guard` is the odd one out — bring it into `cpex-plugins` following that same pattern, then retire the Python copy in the gateway (tracked in IBM/mcp-context-forge#5752).

Reference implementation for structure: `plugins/rust/python-package/pii_filter/`.
Source logic to port: `mcp-context-forge/plugins/output_length_guard/{output_length_guard.py,config.py,guards.py,structured.py}`.

## Scope

- New crate: `plugins/rust/python-package/output_length_guard/` added to the workspace `Cargo.toml` members list.
- Port existing behavior 1:1 first (no silent feature changes):
  - `min_chars`/`max_chars`, `min_tokens`/`max_tokens`, `chars_per_token` estimation
  - `limit_mode` (character/token) and `strategy` (block/truncate)
  - `ellipsis`, `word_boundary` truncation
  - Security limits: `max_text_length`, `max_structure_size`, `max_recursion_depth`
  - Supported shapes: plain str, dict with `text` field, list of strings, MCP content arrays (`[{type: "text", text: ...}]`), and recursive `structuredContent` processing
- PyO3 bindings exposing a `PluginBase`-compatible entry point (`cpex_output_length_guard.OutputLengthGuardPlugin` or equivalent), matching the pii_filter pattern — Rust owns all business logic, Python layer is glue only.
- `plugin-manifest.yaml` carried over/updated (hooks: `tool_post_invoke`, defaults as today).
- Observability: attach a namespaced metrics dict to `result.metadata["output_length_guard"]` when a trace is active (counts of chars/tokens seen, truncated vs blocked, mode/strategy), mirroring the `pii_filter` metrics pattern in `plugins/rust/python-package/pii_filter/src/plugin.rs` (`build_pii_metrics`/`push_metrics_kwarg`). Metrics must carry no raw content, only counts/labels.
- `Makefile`, `pyproject.toml` (package name `cpex-output-length-guard`), `README.md`, maturin build wiring — same shape as `pii_filter`.

## Tests required

- **Unit tests** (Rust `mod tests` inline + Python binding tests under the plugin's own `tests/`): cover all limit modes, both strategies, word-boundary truncation, all supported input shapes, and the security limits (oversized input, deep recursion, huge structures).
- **Plugin-framework integration tests** (`make test-integration` from the crate dir): validate PyO3 entry points load correctly through the Python plugin framework and hook dispatch (`tool_post_invoke`) works end-to-end within `cpex-plugins`, per `TESTING.md`.
- **Coverage**: meet the existing ≥90% Rust line-coverage bar enforced in CI for Rust plugins.
- Note: gateway-level integration tests and E2E tests for this plugin belong in `mcp-context-forge` (see linked gateway issue) per the two-repo testing split in `TESTING.md`.

## Acceptance Criteria

- [ ] New `output_length_guard` Rust crate exists at `plugins/rust/python-package/output_length_guard/`, added to workspace `Cargo.toml`.
- [ ] Feature parity with current Python plugin (all config options, all supported input shapes, all security limits) — no regressions.
- [ ] PyO3 bindings expose an entry point loadable by the gateway's Python plugin framework, matching the `pii_filter`/`secrets_detection` convention.
- [ ] Observability: `result.metadata["output_length_guard"]` metrics emitted when trace is active, no raw content in metrics, matching existing plugin metrics conventions.
- [ ] Unit tests ported/written for all config combinations and edge cases; pass locally via `make test-all`.
- [ ] Plugin-framework integration tests written and passing via `make test-integration`.
- [ ] Rust line coverage ≥90% for the crate.
- [ ] `README.md`, `plugin-manifest.yaml`, `pyproject.toml` (`cpex-output-length-guard`) present and consistent with other migrated plugins.
- [ ] Repo contract tests pass: `python3 -m unittest tests/test_plugin_catalog.py tests/test_install_built_wheel.py` and `python3 tools/plugin_catalog.py validate .`.
- [ ] Initial version tagged/published (e.g. `0.1.0`) so the gateway can pin `cpex-output-length-guard>=0.1.0` (see linked gateway issue for the removal + version bump).
- [ ] Linked to and coordinated with the gateway-side removal issue so both land together (old Python plugin removed only after this package is published and consumable).

## Out of scope

- Removing the Python plugin from `mcp-context-forge` — tracked separately in IBM/mcp-context-forge#5752.
- Adding new functionality beyond current Python behavior (follow-up issue if needed).

prakhar-singh1928:
## Closes #145 — Migrate `output_length_guard` to Rust

Ports the `output_length_guard` plugin from the pure-Python implementation in
`mcp-context-forge` to a Rust core with thin PyO3 bindings, following the
established `pii_filter` pattern. Published as `cpex-output-length-guard`.

Tracked on the gateway side: IBM/mcp-context-forge#5752

---

#### Core migration (`a24a944`)
- New Rust crate at `plugins/rust/python-package/output_length_guard/`
- All Python behaviour ported 1:1 — no silent feature changes
- Rust modules: `config.rs`, `guards.rs`, `structured.rs`, `plugin.rs`
- Thin Python shim at `cpex_output_length_guard/output_length_guard.py` delegates entirely to the Rust core via PyO3
- `plugin-manifest.yaml` — hook: `tool_post_invoke`
- Crate added to workspace `Cargo.toml`
- `tests/test_plugin_catalog.py` updated — plugin count 7→8, all expected plugin lists updated in 5 places
- 21 plugin-framework integration tests added in `plugins/tests/output_length_guard/test_integration.py`
- `uv.lock` committed for reproducible dev installs

#### Bug fixes applied during validation

**`11b4dae` — enforce `max_structure_size` on MCP content lists**
- `process_mcp_items_result` was not checking `max_structure_size` against the content list length; the check only existed inside `process_structured_data` (used for `structuredContent`). Added the guard at the top of `process_mcp_items_result`, matching the existing behaviour in `structured.rs`.
- `MIN_MAX_STRUCTURE_SIZE` lowered from 10 → 1 so that small values (e.g. 2) can be set in config for testing. The unit test updated accordingly (reject 0, not 5).

**`6dff244` — emit observability metrics for MCP `CallToolResult` dicts**
- `handle_mcp_content_dict` was modifying the payload but never calling `push_metrics_kwargs`, so `result.metadata["output_length_guard"]` was silently absent on every traced tool call whose result was a `{"content": [...]}` dict — the most common MCP result shape in production.
- `process_mcp_items_result` return type extended from `(Vec, bool)` to `(Vec, bool, usize, usize)` to surface `total_chars_seen` and `items_modified_count` without a second pass.
- The old `handle_mcp_list` tally was iterating `out_items` (already-truncated dicts) and calling `.extract::<String>()`, which always failed silently, leaving `chars_seen = 0`. Both callers now use the accurate counts from the single processing pass.

---

### Feature parity

All config options and input shapes from the Python original are implemented.

**Config options:** `min_chars`, `max_chars`, `min_tokens`, `max_tokens`, `chars_per_token`, `limit_mode`, `strategy`, `ellipsis`, `word_boundary`, `max_text_length`, `max_structure_size`, `max_recursion_depth`

**Supported input shapes:** plain `str`, `dict` with `text` field, `list[str]`, MCP content array, MCP `CallToolResult` dict with `content` list, `structuredContent` / `structured_content` recursive processing, `type: "resource"` item text field guarding

**Violation codes:** `OUTPUT_LENGTH_VIOLATION`, `OUTPUT_TOKEN_VIOLATION`, `STRUCTURE_SIZE_VIOLATION`, `STRUCTURE_DEPTH_VIOLATION`

**Observability:** `result.metadata["output_length_guard"]` emitted only when `extensions.request.trace_id` is present — counts and labels only, never raw content.

---

### Validation

| Check | Result |
|---|---|
| `cargo clippy -p output_length_guard -- -D warnings` | ✅ PASS |
| `cargo fmt -- --check` | ✅ PASS |
| `cargo test -p output_length_guard` | ✅ 63/63 PASS |
| `make test-integration` | ✅ 21/21 PASS |
| `uv run python3 -m unittest tests.test_plugin_catalog tests.test_install_built_wheel` | ✅ 126/126 PASS (3 skipped, pre-existing) |
| `uv run python3 tools/plugin_catalog.py validate .` | ✅ `{"status": "ok"}` |

---

### Files changed

| Path | Description |
|---|---|
| `plugins/rust/python-package/output_length_guard/` | New plugin crate (Rust + Python shim) |
| `plugins/tests/output_length_guard/test_integration.py` | 21 plugin-framework integration tests |
| `Cargo.toml` | Added new crate to workspace members |
| `Cargo.lock` | Updated automatically |
| `tests/test_plugin_catalog.py` | Plugin count 7→8, all plugin lists updated |

---

### Next steps (gateway side — IBM/mcp-context-forge#5752)

- Install `cpex-output-length-guard>=0.1.0` and update `pyproject.toml` plugins extra
- Update `plugins/config.yaml` `kind` to `cpex_output_length_guard.output_length_guard.OutputLengthGuardPlugin`
- Add `limit_mode`, `strategy` to `_SAFE_STRING_FIELD_NAMES` and `chars_seen`, `truncated_count` to `_SAFE_NUMERIC_FIELD_NAMES` in `mcpgateway/plugins/utils.py`
- Remove `plugins/output_length_guard/` (Python implementation)
- Tag `output-length-guard-v0.1.0` on `main` to trigger PyPI publish


lucarlig:
I compared this PR at `30e41b00324dffe533fc83270cfa9d7909e08cb4` against the current Python `output_length_guard` implementation in `mcp-context-forge` main at `ff1fcc828966e8cf6c1905a4bd189415b6412904`, using both source review and differential calls through the real `cpex` hook models.

There are several observable differences. A difference does **not automatically mean the Rust behavior is bad**—some may be worthwhile security hardening, stricter validation, or intentional cleanup. However, because #145 asks for a 1:1 port with no silent feature changes, could you please review each difference, make a conscious choice about whether it is acceptable, and then either align the implementation or explicitly document and test the intentional divergence?

### Observed differences

1. **UTF-8 slicing can raise a panic**

   `guards.rs::cap_at_max_text_length` slices `&value[..max_text_length]` at an arbitrary byte offset. For example, token mode with `max_tokens=1`, `max_text_length=1000`, and a result of `"€" * 400` returns `"€€€€…"` in Python, while the Rust path raises `pyo3_runtime.PanicException` because byte 1000 splits a character.

2. **Oversized MCP arrays bypass individual length guarding in truncate mode**

   `process_mcp_items_result` returns the oversized list unchanged when `strategy=truncate`, so the item-processing loop below is never reached. With 11 text items, `max_structure_size=10`, and `max_chars=2`, Python truncates each `"abcdef"` to `"a…"`; Rust returns no modified payload. The comment says individual content is still guarded, but the early return prevents that.

   In block mode, Rust also introduces a top-level `STRUCTURE_SIZE_VIOLATION` for MCP content arrays, whereas the Python handler does not apply `max_structure_size` to that top-level MCP array. That hardening may be desirable, but it is a behavioral change to choose explicitly.

3. **Violation transport codes are not preserved**

   The Python implementation sets `mcp_error_code=-32000` and `http_status_code=422` on every violation. Rust `build_violation` does not populate either field. The gateway consumes these fields, so blocked calls can expose a different JSON-RPC code (`-32602` fallback rather than `-32000`), and the generic HTTP exception-handler path can fall back to status 200 because the output-length violation codes are not in `PLUGIN_VIOLATION_CODE_MAPPING`.

4. **Token accounting uses UTF-8 bytes rather than Python Unicode codepoints**

   Rust uses `str::len()` for token estimation, structured token counts, metrics, and some metadata; Python uses `len(str)`. For `"é" * 5`, `chars_per_token=4`, and `max_tokens=1`, Python estimates one token and leaves the result unchanged, while Rust estimates two and truncates it to `"éé…"`. Unicode word-boundary behavior can also differ because some calculations mix byte and character indexes.

5. **Legacy result metadata differs**

   `TextResult::Unchanged` currently conflates within-bounds, numeric-exempt, and below-min-in-truncate-mode states. The caller labels all of them `within_bounds=true`. For a below-min value such as `"short"` with `min_chars=10`, Python returns metadata containing `within_bounds=false`, `limit_mode`, `strategy`, `truncated=false`, and `new_length`; Rust reports `within_bounds=true`.

   Other metadata differences include:

   - blocked Rust results omit the Python length/mode/strategy metadata;
   - `list[str]` results omit Python's per-item `items` metadata and block-mode `violation_index`/`total_items`;
   - numeric strings omit Python's `numeric: true` marker;
   - structured-content block/modify paths omit several Python metadata fields;
   - Rust reports byte lengths for non-ASCII strings where Python reports codepoint lengths.

6. **Configuration validation/defaults differ**

   - Python requires `max_structure_size >= 10`; Rust accepts values down to 1. The Rust integration test using `max_structure_size=2` therefore cannot run against the Python implementation.
   - With an empty plugin config, Python defaults `max_chars` to `None` (unlimited), while Rust defaults it to 15,000 and truncates longer outputs. Both manifests specify 15,000, but direct/plugin-programmatic construction behaves differently.
   - Pydantic accepts/coerces some values such as numeric strings and common boolean forms that the Rust extractor rejects. Stricter Rust parsing may be preferable, but it is another compatibility choice.

7. **Numeric-string exemption is not identical**

   Python uses `math.isfinite(float(text))`; Rust uses `text.trim().parse::<f64>()`. For example, Python treats `"1_000"` as numeric and preserves it, while Rust treats it as non-numeric and can block/truncate it. Some Unicode numeric forms differ as well.

8. **Additional edge-case output differences**

   - An empty list is handled as `list[str]` in Python and returns `metadata={"items": []}`; Rust classifies it as unsupported and returns skipped metadata.
   - Structured dictionaries with non-string keys preserve the data in both implementations, but Rust converts the key to an empty path component for violation reporting, so Python may report location `1` while Rust reports `root`.
   - Unsupported-result metadata differs: Python includes the result type in `reason`, while Rust emits the generic `unsupported_type`.

### Verification performed

- Rust PR suite: 141 Rust unit tests passed.
- Rust/PyO3 integration suite: 21 tests passed.
- Python reference suite: 326 tests passed.
- Current PR CI is green.

Those suites validate each implementation independently, but they do not currently assert cross-language equivalence. A small table-driven differential suite that invokes both implementations with the same configs/payloads would make the intended compatibility decisions durable. Gateway-level E2E coverage is still tracked separately in IBM/mcp-context-forge#5752.

Again, I am not assuming every difference should be removed. Please make and record the intended choice for each one so consumers know which behaviors are compatibility guarantees and which are intentional changes.

plugins/rust/python-package/output_length_guard/src/guards.rs
```diff
@@ -0,0 +1,960 @@
...
+/// Returns a sub-slice capped at `max_text_length` bytes.
+///
+/// Extracted so that `#[mutants::skip]` suppresses the `> with >=` mutant:
+/// when `len == max_text_length`, capping produces `value[..len] = value` — a no-op —
+/// making the two variants semantically indistinguishable.
+#[mutants::skip] // equivalent: > vs >= when len == max_text_length; capping to self = no-op
+#[inline]
+fn cap_at_max_text_length(value: &str, max_text_length: usize) -> &str {
+    if value.len() > max_text_length {
+        &value[..max_text_length]
```
lucarlig: [P1] This slices at an arbitrary UTF-8 byte offset. With token mode, `max_tokens=1`, `max_text_length=1000`, and `"€" * 400`, Python returns `"€€€€…"`, while Rust raises `pyo3_runtime.PanicException` because byte 1000 splits a character. Please use a codepoint/character-boundary-aware cap. This looks like a correctness issue rather than merely a compatibility preference.

plugins/rust/python-package/output_length_guard/src/plugin.rs
```diff
@@ -0,0 +1,2083 @@
...
+                            "max_size".to_string(),
+                            serde_json::json!(self.cfg.max_structure_size),
+                        ),
+                    ],
+                )?;
+                return Ok(Err(violation));
+            }
+            // Truncate strategy: pass the oversized list through unchanged (items will
+            // still have their individual text content guarded below).
+            return Ok(Ok((
```
lucarlig: [P1] In truncate mode this early return passes the entire oversized MCP array through unchanged, so the item-processing loop below never runs despite the comment. With 11 items, `max_structure_size=10`, and `max_chars=2`, Python truncates every `abcdef` to `a…`; Rust returns no modified payload. Please either continue into individual guarding or explicitly choose and document this bypass behavior.

plugins/rust/python-package/output_length_guard/src/plugin.rs
```diff
@@ -0,0 +1,2083 @@
...
+    ) -> PyResult<Result<(Vec<Py<PyAny>>, bool, usize, usize), Py<PyAny>>> {
+        // Security: enforce max_structure_size regardless of strategy.
+        // Mirrors the pattern in structured.rs::process_list / process_dict.
+        if list.len() > self.cfg.max_structure_size {
+            log::error!(
+                "Content list size {} exceeds maximum {} (MCP items)",
+                list.len(),
+                self.cfg.max_structure_size
+            );
+            if self.cfg.strategy == Strategy::Block {
```
lucarlig: [P2] Applying `max_structure_size` to a top-level MCP content array in block mode is new behavior: the Python MCP-array handlers do not enforce this structural limit there. The hardening may be desirable, but it is not a 1:1 port. Please consciously retain and document/test it, or align with the Python scope.

plugins/rust/python-package/output_length_guard/src/plugin.rs
```diff
@@ -0,0 +1,2083 @@
...
+        py,
+        "PluginViolation",
+        vec![
+            ("reason", reason.into_pyobject(py)?.into_any().unbind()),
+            (
+                "description",
+                description.into_pyobject(py)?.into_any().unbind(),
+            ),
+            ("code", code.into_pyobject(py)?.into_any().unbind()),
+            ("details", details_dict.into_any().unbind()),
```
lucarlig, line 765: [P1] The Python implementation sets `mcp_error_code=-32000` and `http_status_code=422` on every violation, but this builder omits both. The gateway consumes these fields, so blocked requests can change to JSON-RPC `-32602`; the generic HTTP handler can also fall back to status 200 because these violation codes are not mapped. Please preserve the Python transport codes unless this client-visible change is explicitly accepted.

plugins/rust/python-package/output_length_guard/src/guards.rs
```diff
@@ -0,0 +1,960 @@
...
+            let above_max = cfg.max_tokens.is_some_and(|max| token_count > max);
+            (below_min, above_max)
+        }
+    }
+}
+
+/// Estimate token count using configurable chars-per-token ratio.
+pub fn estimate_tokens(text: &str, chars_per_token: usize) -> usize {
+    let cpt = chars_per_token.max(1);
+    text.len() / cpt
```
lucarlig: [P2] `str::len()` counts UTF-8 bytes, while Python `len(str)` counts Unicode codepoints. For `"é" * 5`, `chars_per_token=4`, and `max_tokens=1`, Python estimates one token and leaves the value unchanged; Rust estimates two and truncates it. The same choice appears in structured token counting and metrics. Byte counting may be intentional, but please choose one semantic consistently and document/test the divergence if retained.

plugins/rust/python-package/output_length_guard/src/plugin.rs
```diff
@@ -0,0 +1,2083 @@
...
+                    build_text_meta_dict(py, text, &new_text_str(&new_result_obj, py), false)?;
+                merge_metrics_into_meta(py, &meta, trace_id, text.len(), true, 1, &self.cfg)?;
+                let kwargs: Vec<(&str, Py<PyAny>)> = vec![
+                    ("modified_payload", new_payload),
+                    ("metadata", meta.into_any().unbind()),
+                ];
+                build_result_dyn(py, "ToolPostInvokeResult", kwargs)
+            }
+            TextResult::Unchanged => {
+                let meta = build_text_meta_dict(py, text, text, true)?;
```
lucarlig: [P2] `TextResult::Unchanged` conflates within-bounds, numeric-exempt, and below-min-in-truncate-mode states, yet this branch labels all of them `within_bounds=true`. Python reports a below-min value as `within_bounds=false` with `limit_mode`, `strategy`, `truncated=false`, and `new_length`. Please carry enough state/metadata to preserve the old contract, or explicitly approve and document the metadata change.

plugins/rust/python-package/output_length_guard/src/plugin.rs
```diff
@@ -0,0 +1,2083 @@
...
+                chars_seen: 0,
+                truncated_count: 0,
+                blocked: true,
+                mode: cfg.limit_mode.as_str(),
+                strategy: cfg.strategy.as_str(),
+                stage: "tool_post_invoke",
+            },
+        )
+    {
+        kwargs.push(("metadata", md.into_any().unbind()));
```
lucarlig: [P2] Without a trace ID, blocked Rust results have no metadata; with a trace ID, they contain only the new metrics namespace. Python always returns the original length/mode/strategy metadata for the offending text, and structured-content blocks add their structured location/token fields. Please merge observability metrics into the legacy metadata rather than replacing that contract, unless dropping it is consciously accepted.

plugins/rust/python-package/output_length_guard/src/plugin.rs
```diff
@@ -0,0 +1,2083 @@
...
+                items_modified,
+                &self.cfg,
+            )?;
+            let kwargs: Vec<(&str, Py<PyAny>)> = vec![
+                ("modified_payload", new_payload),
+                ("metadata", meta.into_any().unbind()),
+            ];
+            return build_result_dyn(py, "ToolPostInvokeResult", kwargs);
+        }
+        let meta = PyDict::new(py);
```
lucarlig, line 315: [P2] Python `list[str]` handling returns per-item metadata under `items`; on a block it also returns `violation_index` and `total_items`. This branch returns an empty dict (plus optional trace metrics), so consumers lose that audit information. Please preserve it or document/test the intentional metadata break.

plugins/rust/python-package/output_length_guard/src/plugin.rs
```diff
@@ -0,0 +1,2083 @@
...
+            }
+        }
+
+        if struct_modified {
+            let nd = new_result_dict_data.unwrap();
+            let new_payload = clone_payload_with_attr(py, payload, "result", &nd)?;
+            let meta = PyDict::new(py);
+            meta.set_item("mcp_result_processed", true)?;
+            meta.set_item("items_modified", true)?;
+            meta.set_item("structured_content_processed", true)?;
```
lucarlig, line 380: [P2] The Python structured-content modified path also includes `min_tokens`, `max_tokens`, and `chars_per_token`; the Rust result stops at these three boolean/status fields. This may be metadata simplification rather than faulty enforcement, but it is observable behavior. Please explicitly choose whether those fields remain part of the compatibility contract.

plugins/rust/python-package/output_length_guard/src/config.rs
```diff
@@ -0,0 +1,552 @@
...
+use pyo3::prelude::*;
+use pyo3::types::{PyAny, PyDict};
+use thiserror::Error;
+
+// Security limit constants (match Python config.py validators)
+pub const MIN_MAX_TEXT_LENGTH: usize = 1_000;
+pub const MAX_MAX_TEXT_LENGTH: usize = 10_000_000;
+pub const DEFAULT_MAX_TEXT_LENGTH: usize = 1_000_000;
+
+pub const MIN_MAX_STRUCTURE_SIZE: usize = 1;
```
lucarlig, line 17: [P2] Python validates `max_structure_size` in `[10, 100000]`; Rust accepts values down to 1, and the new integration test relies on the Python-invalid value 2. Stricter testability/security tuning may justify this change, but please do not present it as identical behavior: either restore 10 or document and test the new supported range.

plugins/rust/python-package/output_length_guard/src/config.rs
```diff
@@ -0,0 +1,552 @@
...
+pub const MIN_MAX_RECURSION_DEPTH: usize = 10;
+pub const MAX_MAX_RECURSION_DEPTH: usize = 1_000;
+pub const DEFAULT_MAX_RECURSION_DEPTH: usize = 100;
+
+pub const DEFAULT_CHARS_PER_TOKEN: usize = 4;
+pub const MIN_CHARS_PER_TOKEN: usize = 1;
+pub const MAX_CHARS_PER_TOKEN: usize = 10;
+
+pub const DEFAULT_ELLIPSIS: &str = "\u{2026}"; // …
+pub const DEFAULT_MAX_CHARS: Option<usize> = Some(15_000);
```
lucarlig: [P2] With an empty plugin config, the Python model defaults `max_chars` to `None` (unlimited), while Rust defaults to 15,000 and truncates longer results. The manifests both advertise 15,000, so the desired behavior needs a conscious decision: align direct construction, or document that Rust now applies manifest-like defaults internally.

plugins/rust/python-package/output_length_guard/src/config.rs
```diff
@@ -0,0 +1,552 @@
...
+            }
+            cfg.min_tokens = n as usize;
+        }
+
+        if let Some(resolved) = Self::parse_optional_usize(dict, "max_tokens")? {
+            cfg.max_tokens = resolved;
+        }
+
+        if let Some(val) = dict.get_item("chars_per_token")? {
+            let n: usize = val.extract()?;
```
lucarlig, line 210: [P2] Direct PyO3 extraction is stricter than the Pydantic configuration boundary it replaces. The Python model accepts/coerces values such as `max_chars="10"`, `chars_per_token="4"`, and common boolean forms, while Rust raises `TypeError`. Strict parsing may be preferable, but please decide whether this is an intentional configuration-compatibility break and document/test it if retained.

plugins/rust/python-package/output_length_guard/src/guards.rs
```diff
@@ -0,0 +1,960 @@
...
+}
+
+/// Check if a string represents a finite numeric value.
+/// Handles integers, floats, and scientific notation.
+/// Rejects nan, inf, and strings longer than 50 chars.
+pub fn is_numeric_string(text: &str) -> bool {
+    if text.len() > MAX_NUMERIC_STRING_LENGTH {
+        return false;
+    }
+    match text.trim().parse::<f64>() {
```
lucarlig, line 219: [P2] Rust `f64` parsing does not recognize exactly the same numeric strings as Python `float()`. For example, Python treats `"1_000"` as finite numeric content and exempts it, while Rust treats it as non-numeric and can block/truncate it; some Unicode numeric forms also differ. Neither parser is inherently the only correct choice, but please consciously define the numeric-exemption grammar and add parity/intent tests.

plugins/rust/python-package/output_length_guard/src/plugin.rs
```diff
@@ -0,0 +1,2083 @@
...
+        }
+
+        // Case 1: Plain string
+        if let Ok(text) = result_val.extract::<String>() {
+            return self.handle_plain_string(py, payload, &text, trace_id.as_deref(), &name);
+        }
+
+        // Case 3 & 4: List
+        if let Ok(list) = result_val.cast::<PyList>()
+            && !list.is_empty()
```
lucarlig, line 79: [P3] This non-empty gate changes the empty-list path. Python's `all(isinstance(...))` treats `[]` as `list[str]` and returns `metadata={"items": []}`; Rust classifies it as unsupported/skipped. Please align or explicitly accept and test the new classification.

plugins/rust/python-package/output_length_guard/src/structured.rs
```diff
@@ -0,0 +1,976 @@
...
+            value: dict.clone().into_any().unbind(),
+            modified: false,
+        });
+    }
+
+    let mut modified = false;
+    let out_dict = PyDict::new(py);
+
+    for (key, value) in dict.iter() {
+        let key_str = key.extract::<String>().unwrap_or_default();
```
lucarlig: [P3] Non-string dict keys become an empty path component because extraction falls back to `""`. Python string-formats the key, so a violation under `{1: "abcdef"}` reports location `1`, whereas Rust reports `root`. The data remains intact, but violation diagnostics differ; please stringify non-string keys or document the narrower key contract.

plugins/rust/python-package/output_length_guard/src/plugin.rs
```diff
@@ -0,0 +1,2083 @@
...
+            // Case 4: List of strings
+            if list.iter().all(|item| item.extract::<String>().is_ok()) {
+                return self.handle_string_list(py, payload, list, trace_id.as_deref(), &name);
+            }
+        }
+
+        // Unsupported result type
+        let meta = PyDict::new(py);
+        meta.set_item("skipped", true)?;
+        meta.set_item("reason", "unsupported_type")?;
```
lucarlig: [P3] Python emits `reason=unsupported_type_<type name>`, while Rust emits only `unsupported_type`. This is a small metadata difference, but downstream diagnostics can observe it. Please retain the type suffix or explicitly accept/test the simplified reason.

lucarlig (changes requested):
Requesting changes because the PR currently claims a 1:1 behavioral port, while the inline threads identify observable compatibility differences. A difference is not automatically bad: some may be desirable hardening or cleanup. Please make a conscious choice for each thread—either align with the Python counterpart, or explicitly document and regression-test the intentional divergence. The earlier consolidated comment contains the differential-test matrix and verification details.

prakhar-singh1928:
Thanks for the detailed review, Luca. All four blocking issues are resolved. Here's the full picture.

──────────────────────────────────────────────────────────────────────
FIXED — Issues 1–4
──────────────────────────────────────────────────────────────────────

Issue 1 — UTF-8 panic in cap_at_max_text_length
The old &value[..max_text_length] slice is replaced with a snap loop:

    while cut > 0 && !value.is_char_boundary(cut) { cut -= 1; }

Regression: cap_at_max_text_length_does_not_panic_on_multibyte_boundary
("€".repeat(400), max_text_length=1000 — byte 1000 falls mid-codepoint;
snap moves it to 999).

Issue 2 — Oversized MCP content list bypassed per-item guarding in truncate mode
The STRUCTURE_SIZE_VIOLATION early-return in process_mcp_items_result is now
gated on strategy == Block only. In truncate mode a warning is logged and the
item loop proceeds so each string is still truncated individually.

Note: the block-mode top-level STRUCTURE_SIZE_VIOLATION for MCP content arrays
is an intentional hardening divergence from the Python implementation (which
does not apply max_structure_size to that array). It is explicitly tested and
kept as a deliberate security choice.

Regressions:
- mcp_content_dict_oversized_list_truncate_mode_still_guards_item_text
  (3 items, max_structure_size=2, max_chars=2: each item text verified ≤ 2 chars)

Issue 3 — Violation transport codes not populated
build_violation now passes mcp_error_code=-32000 and http_status_code=422 on
every PluginViolation, satisfying the gateway contract and avoiding the
-32603 fallback in the exception handler.

Regression: violation_carries_mcp_error_code_and_http_status_code
(asserts both fields equal -32000 / 422 on the returned violation object).

Issue 4 — below_min in truncate mode returned within_bounds=true
TextResult::BelowMin is now a distinct enum variant. handle_text returns it
when below_min && !above_max in truncate mode. handle_plain_string and
handle_text_dict have a dedicated BelowMin arm calling
build_text_meta_dict(..., false) (within_bounds=false). The three other
match sites use BelowMin | Unchanged => and fall through identically.

Regression: below_min_truncate_mode_reports_within_bounds_false
(min_chars=20, input "short": within_bounds=false, no modified_payload).

──────────────────────────────────────────────────────────────────────
INTENTIONAL DIVERGENCES — Issues 5–8 (not changed in this PR)
──────────────────────────────────────────────────────────────────────

These are acknowledged differences that are either deliberate hardening or
out of scope for this PR. Each is explicitly tested so the behaviour is durable.

5 — Token estimation uses chars().count() not byte len()
   Rust uses char_count (Unicode codepoints) for limit evaluation and metadata,
   matching the intent of max_chars/min_chars. The byte-count path in
   estimate_tokens is a known approximation in both implementations.
   Will track alignment in a follow-up.

6 — Several result metadata fields absent
   (blocked paths, list items, numeric marker, structured-content fields,
   byte vs codepoint lengths in metadata)
   These are observability-only fields that do not affect gateway routing,
   violation codes, or payload modification. Tracked as a follow-up against
   the integration test suite.

7 — Config validation / defaults differ
   Stricter Rust parsing is preferred over silent Pydantic coercion. The
   max_structure_size >= 10 Python minimum is a heuristic guard, not a
   security requirement. The empty-config max_chars default difference is
   a manifest-layer concern — both manifests specify 15000. Documented.

8 — "1_000" is non-numeric in Rust (trim().parse::<f64>()) but numeric in Python
   Python's float("1_000") accepting underscore separators is a language quirk.
   Rust's stricter parse is preferable — underscore-separated numbers are not
   a common MCP output format and should not be exempt from length guarding.
   Documented.

plugins/rust/python-package/output_length_guard/src/plugin.rs
prakhar-singh1928: Resolved in plugin.rs.

Blocked results now always preserve legacy metadata, regardless of whether a trace ID is present. The metadata includes within_bounds, limit_mode, strategy, truncated, violation details, and original_length/new_length when available. Structured-content violations also preserve fields such as location, token_count, size, and security-limit values.

When a trace ID is present, the output_length_guard observability namespace is merged into this existing metadata rather than replacing it. Added regression coverage for plain-string and structured-content blocks.

plugins/rust/python-package/output_length_guard/src/plugin.rs
prakhar-singh1928: Resolved in plugin.rs.

String-list handling now preserves the legacy audit metadata:

metadata["items"] contains per-item metadata.
Blocked lists include violation_index and total_items.
Items processed before the violation remain available under metadata["items"].
Trace metrics continue to be merged into the same metadata dictionary.
Added regression tests covering per-item metadata in truncate mode and violation position/count in block mode.

plugins/rust/python-package/output_length_guard/src/config.rs
prakhar-singh1928: Intentional compatibility difference: Rust accepts max_structure_size values from 1 to 100,000, while the Python implementation accepts 10 to 100,000. The lower Rust minimum is deliberate because it supports tighter per-result structure limits and enables focused tests with small collections. This behavior is covered by Rust configuration tests, including acceptance of the minimum value 1.

plugins/rust/python-package/output_length_guard/src/plugin.rs
prakhar-singh1928: Resolved by retaining these fields as part of the compatibility contract.

The Rust structured-content modified path now includes:

min_tokens
max_tokens
chars_per_token
This preserves the Python implementation’s observable metadata behavior. Regression coverage confirms the configured values are present after structured-content modification.

plugins/rust/python-package/output_length_guard/src/config.rs
prakhar-singh1928: Intentional compatibility difference: Rust uses strict typed extraction for plugin configuration rather than Pydantic-style coercion. Values such as max_chars: "10", chars_per_token: "4", or string boolean forms are rejected with TypeError instead of being silently converted. This is deliberate to fail fast on malformed configuration and avoid ambiguous runtime behavior; valid configuration values must use their declared numeric and boolean types.

plugins/rust/python-package/output_length_guard/src/guards.rs
prakhar-singh1928: Intentional compatibility difference: Rust defines numeric-string exemptions using trim().parse::<f64>() and requires the parsed value to be finite. Therefore, only standard Rust f64 numeric syntax is exempt; underscore-separated values such as "1_000" and Unicode-specific numeric forms are treated as ordinary text and remain subject to length enforcement.

This is deliberate because accepting language-specific formatting would make the exemption grammar inconsistent and could allow values to bypass output-length limits unexpectedly. The Rust behavior is covered by numeric-string tests for integers, floats, scientific notation, non-numeric values, non-finite values, and the maximum length boundary.

plugins/rust/python-package/output_length_guard/src/plugin.rs
prakhar-singh1928: Intentional compatibility difference: Rust requires a non-empty list before classifying a result as list[str]. Empty lists are therefore treated as unsupported and return skipped metadata, rather than items: [] as in the Python implementation. This avoids inferring an element type from an empty collection. The behavior should be covered by an explicit empty-list regression test.

plugins/rust/python-package/output_length_guard/src/plugin.rs
prakhar-singh1928: Resolved. Rust now includes the unsupported result type in the metadata reason, matching Python’s unsupported_type_<type name> contract. Added regression coverage for unsupported result types.

plugins/rust/python-package/output_length_guard/src/structured.rs
prakhar-singh1928: Resolved. Non-string dictionary keys are now converted to their string representation when constructing structured-content paths, matching the Python implementation. For example, a violation under {1: "abcdef"} reports location 1 instead of root. The underlying dictionary data remains unchanged, and regression coverage was added for non-string keys.

plugins/rust/python-package/output_length_guard/src/plugin.rs
lucarlig: Rechecked at `72fc3ef821c889e66f02578d8abd981b06acf5b8` through the real cpex hook models against ContextForge main (`b13ebe48c98054426ec3ce1e2b37a94fcd0aa194`). This is partially fixed, but two list audit cases still differ:

- With `min_chars=10`, `max_chars=None`, `strategy="truncate"`, and result `["short"]`, Rust returns `items[0].within_bounds=true`; Python returns `false` plus `limit_mode`, `strategy`, `truncated=false`, and `new_length=5`. The `BelowMin | Unchanged` arm at lines 279–281 still merges these states. Please give `BelowMin` its own metadata path.
- With `max_chars=5`, `strategy="block"`, and `["ok", "abcdef", "later"]`, both return `violation_index=1`, but Rust's `items` has length 1 while Python's has length 2, including the offending item. Thus `items[violation_index]` is absent in Rust. Please append the offending item's metadata before returning the block (the current regression asserts the incompatible length of 1).

A difference need not be bad, but the response says this contract is restored. Please finish these cases or explicitly describe and test the intended replacement contract.

plugins/rust/python-package/output_length_guard/src/plugin.rs
lucarlig: Rechecked at `72fc3ef821c889e66f02578d8abd981b06acf5b8`. Plain-string blocked metadata and merging trace metrics are fixed. The structured-content block path is still incomplete:

For `max_chars=5`, `strategy="block"`, and `{"content":[{"type":"text","text":"ok"}],"structuredContent":{"answer":"abcdef"}}`, Python returns `structured_content_blocked=true`, `location="structuredContent"`, `min_tokens=0`, `max_tokens=None`, and `chars_per_token=4`. Rust omits the structured-block marker and token configuration, and sets metadata `location="answer"` from violation details. The new test checks the leaf path but does not verify the legacy structured-block contract.

Please preserve these fields at the structured-content caller before merging metrics. Keep the leaf path in `violation.details.location` (or a separately named metadata field) if the legacy metadata location remains part of the contract. Alternatively, explicitly document and test the changed schema; this is still observable behavior, not a completed parity fix.

plugins/rust/python-package/output_length_guard/src/guards.rs
```diff
@@ -0,0 +1,994 @@
...
+    if value.is_empty() || cut == 0 {
+        return cut;
+    }
+    let cut = cut.min(value.len());
+    let search_back = (max_chars as f64 * 0.2) as usize;
+    let min_search = cut.saturating_sub(search_back);
+
+    // Walk backwards from cut-1 down to min_search
+    let chars: Vec<char> = value[..cut].chars().collect();
+    for i in (min_search..chars.len()).rev() {
```
lucarlig: [P2] Word-boundary search still mixes byte offsets and character indexes. With `max_chars=10`, `word_boundary=true`, default ellipsis, and `"é" * 7 + " abcdefgh"`, Python returns `"ééééééé …"`, while this Rust hook returns `"ééééééé a…"`, cutting into the next word. Here `cut` is byte offset 16, so `min_search` is 14, but the collected prefix has only 9 characters; the loop is empty and never sees the space. Please compute the search window in Unicode codepoint indexes and convert only the chosen boundary back to a byte offset. Add a multibyte prefix regression asserting the exact output. This remains after the token-count fix and affects the advertised `word_boundary` option.

lucarlig (commented):
Re-reviewed the latest head against ContextForge main. Fresh verification: 158 Rust unit tests and 23 plugin-framework integration tests passed; current CI is green. Direct calls confirm the earlier panic, MCP truncate bypass, transport codes, ordinary Unicode token counting, plain below-min metadata, empty-config default, non-string key diagnostics, and unsupported-type reason fixes. The structured-content modified-path token metadata is also restored, so I am resolving that thread.

Two existing metadata threads remain partially addressed; I have added exact reproductions there. Unicode word-boundary truncation also still mixes byte and codepoint indexes (inline reproduction below). These findings prevent me from confirming all review concerns are addressed or approving this head. The explicitly chosen configuration/numeric/empty-list differences remain compatibility decisions rather than automatic defects. Gateway E2E validation remains separate from the plugin-level tests run here.

plugins/rust/python-package/output_length_guard/src/guards.rs
prakhar-singh1928: Fixed : Unicode word-boundary truncation when the input contains multibyte characters. The search window now uses Unicode codepoint indexes and converts only the selected boundary back to a byte offset.

prakhar-singh1928:
@lucarlig for metadata changes
- Fixed list[str] metadata for below-minimum values:
Reports `within_bounds: false`
Preserves `limit_mode` and `strategy`
Includes `truncated: false` and `new_length`
- Fixed blocked `list[str] `metadata to include the offending item under `metadata["items"]` ,  along with the existing `violation_index` and `total_items`.
- Restored structured-content blocked metadata:
```
structured_content_blocked: true
location: "structuredContent"
min_tokens
max_tokens
chars_per_token
```
- Added regression coverage for all four metadata cases and the Unicode word-boundary case.

lucarlig (approved):
LGTM

